PR #9, opened 2015-01-01
Custom HTML, github fi links, re-searching when backend drop-down is changed, permalinks for backends with periods in them

philz:
The four commits below are mutually independent, so feel free to pick and choose!  See the commit messages for more detailed descriptions.

I've been testing manually.


nelhage:
Awesome, these all look good to me.

PR #15, opened 2015-09-17
Fix some subtle index key bugs.

nelhage:
reported by @shalecraig, this manifests as an incorrect index-key generation for case-insensitive matches again `(scope|permission)`


shalecraig:
\o/

PR #18, opened 2015-11-18
Correctly render OpenSearch template

bkrausz:
r? @nelhage 

XML doesn't render properly under html/template, resulting in the `&lt;` at http://livegrep.com/opensearch.xml

Not the cleanest solution, but without pulling in https://github.com/tcard/go-template-interface or similar I don't think we can do much better.

That said, this is my first time writing Go, so :dogscience:

### DIFF
--- a/server/templates.go
+++ b/server/templates.go
@@ -6,6 +6,7 @@ import (
 	"log"
 
 	"html/template"
+	texttemplate "text/template"
 
 	"github.com/livegrep/livegrep/server/config"
 )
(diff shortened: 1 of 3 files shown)


server/templates.go
```diff
@@ -16,6 +17,14 @@ type page struct {
 	Config *config.Config
 }
 
+func executeTextTemplate(t *texttemplate.Template, context interface{}) ([]byte, error) {
```
nelhage: Let's reduce this duplication a little by

```
type Template {
  Execute(wr io.Writer, data interface{}) error
}
```

and making `executeTemplate` take a `Template`

nelhage:
otherwise seems reasonable to me.

bkrausz:
Updated as requested

nelhage:
:+1:

PR #19, opened 2016-01-09
Update google analytics tracking code

tpcwang:
Updated script according to this page:
https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs


nelhage:
:+1: thanks

Issue #21, opened 2016-01-12
Feature request: add repo support for other VCS through fs_paths


tpcwang:
My use case is that I want to index source code on various VCS with other source browsers that I have already set up. I don't think livegrep should add support for those VCS, but it would be nice to indirectly support them through the file indexer. I think the following tasks will be necessary:
- [x] Expand fspaths from arrays of strings to arrays of objects with similar structure as repositories
- [x] Support rendering links using a custom format (possibly plumbing metadata through the layers)

What are your thoughts? I'll look to see if I can implement it if you are okay with this approach.

nelhage:
Yeah, I'd be happy to have this feature, and to take a PR of that form. Some sort of generic plumbing of custom formats through from the config to the web interface would be awesome, but something more hackish that supported keys other than `github` in the metadata would also be fine for now.

PR #22, opened 2016-01-13
Add repo support for other VCS through fs_paths

tpcwang:
Addresses #21 

This does break existing config files containing fs_paths - not sure if that's okay. I tried to maintain compatibility for the github metadata, which I figured is probably more important. Also, maybe there should be a better name other than fs_paths.


nelhage:
Awesome, this looks great. I'd agree with your choices about keeping compatibility.

PR #24, opened 2016-01-25
Simplified callback declaration and added an optional transform function

tpcwang:
(1) Use boost::function to simplify the callback function declaration
(2) Allow a filter/transform function to be specified. I have a prototype to look at an index of tags files and then transform the tags match to the actual code match. I can remove this second part if you'd prefer, since it may stay unused if my ctags prototype doesn't make it.

### DIFF
--- a/src/codesearch.cc
+++ b/src/codesearch.cc
@@ -886,9 +886,9 @@ code_searcher::search_thread::search_thread(code_searcher *cs)
     }
 }
 
-void code_searcher::search_thread::match_internal(const query &q,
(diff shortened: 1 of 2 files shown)


nelhage:
The `boost::function` simplification is a nice cleanup. I think I'd rather hold (2) until there's a PR that also uses it, instead of leaving it sitting around unused for now, if you don't mind.

tpcwang:
Reverted (2) and left only the boost::function cleanup.

Issue #23, opened 2016-01-21
Pie-in-the-sky feature request: searching for definitions using ctags results


tpcwang:
Let me first say that this possibly falls outside of what livegrep should reasonably do, but I'll throw it out there and see what you think.

It is great that livegrep can search for any text in the code, just like plain old grep, but I often have a symbol and I just want to go to the definition quickly. This can be troublesome for symbols that are common, such as [open](https://livegrep.com/search/linux?q=open). It is arguable that the user should craft a better regex to narrow down the search, such as [int.*open\\( file:\\.c](https://livegrep.com/search/linux?q=int.*open%5C\(+file%3A%5C.c), but I wonder if we can do better than that.

One possibility is to implement an indexer that uses the output of ctags and only add the tag name to the suffix array instead of entire lines. If we can maintain the line context, then users who want to search for symbols can search for it in the "tags" repo and the results will only contain definitions.

Another possibility is to implement a different backend that implements its own mechanism using the tags file for returning the results to the frontend.

Is this outside the scope of livegrep?

tpcwang:
Looking through the code, I think possibility 1 will muddy the implementation too much due to the way that the line context and the match results are constructed. Feel free to close this issue if you think possibility 2 doesn't belong in livegrep.

nelhage:
Yeah, I suspect this doesn't make sense as a baked-in thing in livegrep. I'd potentially be interested in frontend support for plugging into some other backend that runs side-by-side with the backend, but I don't have any particular thoughts on the shape of that.

tpcwang:
Fair enough. Thanks for the consideration!

PR #26, opened 2016-01-29
Added support for searching through ctags results

tpcwang:
This addresses what I wanted in #23. 

To use this, you would:
1. Create a tags file with ctags on your source
2. Run codesearch -dump_index on the tags file
3. Run codesearch -load_index and -load_tags, with the tags backend listening on another port
4. Make sure to modify server.json to refer to both backends

The overall algorithm is pretty simple. I use the tags index for searching through lines in the tags file. For each match in the tags file, I transform the match by using the source index to get the context lines.

There are still some rough spots, such as lack of documentation on setting this up, or the fact that you need to ensure that the tags file would have the same path as the path stored in the index file. However, I find that this is actually quite useful in finding definitions quickly, while still maintaining the speed that livegrep is known for.

### DIFF
--- a/src/tagsearch.cc
+++ b/src/tagsearch.cc
@@ -13,30 +13,9 @@
 
 using re2::RE2;
 
-void tag_searcher::load_index(const string& path) {
-    cs_.load_index(path);
-}
+namespace {
 
-void tag_searcher::cache_indexed_files(code_searcher* cs) {
-    file_alloc_ = cs->alloc_;
-    for (auto it = cs->begin_files(); it != cs->end_files(); ++it) {
-        auto file = *it;
-        auto key = repo_path_pair(file->tree->name, file->path);
-        path_to_file_map_.insert(std::make_pair(key, file));
-    }
-}
-
-std::string tag_searcher::create_tag_line_regex(
-    const std::string& name,
-    const std::string& file,
-    const std::string& lno,
-    const std::string& tags) const
-{
-    // full regex match for a tag line created with ctags using format=2.
-    return std::string("^") + name + "\\t" + file + "\\t" + lno + "\\;\\\"\\t" + tags + "$";
-}
-
-std::string tag_searcher::create_partial_regex(RE2 *re) const {
+std::string create_partial_regex(RE2 *re) {
     if (!re)
         return ".*";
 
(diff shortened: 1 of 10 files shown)


nelhage:
Sorry, this is on my list to look at. One question, at a glance, though: would it be possible to do the regex-futzing and conversion outside of the `codesearch` index entirely? I'm not opposed to taking it into the backend, but if it's really just a wrapper around the normal index, I wonder if we can move it out of process entirely, for simplicity.

I'll take a more detailed look when I have time – sorry if that's an obviously-silly suggestion once I look more closely.

tpcwang:
No worries. There is no rush to get this merged in. I also want to make sure that this is something that would be useful to others, and that it goes through careful review.

I actually started working on this backend in another process, but I moved it in for performance reasons. The tagsearch service is merged with the codesearch backend because it uses the codesearch backend's cache of file contents to provide context information to the user. One could argue that tagsearch can implement its own context lookup, but this seems like a terrible waste of system resource for tagsearch to implement its own caching or memory mapping if I want to run both the tagsearch and codesearch backends.

src/tagsearch.cc
```diff
@@ -0,0 +1,118 @@
...
+    const std::string& name,
+    const std::string& file,
+    const std::string& lno,
+    const std::string& tags) const
+{
+    // full regex match for a tag line created with ctags using format=2.
+    return std::string("^") + name + "\\t" + file + "\\t" + lno + "\\;\\\"\\t" + tags + "$";
+}
+
+std::string tag_searcher::create_partial_regex(RE2 *re) const {
```
nelhage: I'd rather make these helpers normal functions inside an anonymous namespace, as opposed to methods, I think, since they don't actually act on instances.

nelhage:
Hm, this is pretty cute. I do wonder if `codesearch` should support querying for file contents or file ranges, and whether that would let us pull this out into an external client, but I'd tentatively be up for taking this patch.

Thoughts on whether you'd like to carry this in `master` as-is, or what next steps are? Have you been using this successfully?

tpcwang:
If `codesearch` supports querying for file contents, then I think we would be closer to pulling this out into an external client. However, adding support for negation filters for this external client would be more complicated because I don't think you can construct a single regex that will result in the negation filters being applied. If this external client applies the negation filter after retrieving the result from `codesearch`, then it may not be returning all possible results. The current implementation works around this through the transform method, which can bypass `codesearch`'s internal match count.

I would certainly like to carry this in master if you think this feature is useful (after I address the comment that you have). I think the next step would be to make this feature easy for others to use. I personally use docker-compose to sync down the source, generate the index and tag files, and bring up the services in a consistent manner, but anything that can automate the generation of tags and tags index file will be a big plus.

I spend about equal time using the "grep" search and "ctags" search. ctags search is really useful when I want to find definitions quickly, especially so if there are a lot of references or if you have a lot of different repos. For example, if I want to look at where termios is defined, this may be my searches using grep search:
1. https://livegrep.com/search/linux?q=termios (and see that there are too many references)
2. https://livegrep.com/search/linux?q=struct+termios (and possibly miss it because it's 8th on the list)
3. https://livegrep.com/search/linux?q=struct+termios+%7B (if I happen to know the Linux coding convention)

With ctags search, I simply have to search for termios (or ^termios$) and skip step 1&2.

nelhage:
Cool, I'm happy to carry this, with the caveat that if we figure out a better way to do it I might make or take breaking refactors (I'll try to keep you in the loop, if so!). I'll also consider adding tags support to the public livegrep.com to make sure I get this code exercised.

tpcwang:
Sounds good to me. I'm not particularly attached to ctags or my approach, so anything that can help me find definitions easily would work for me.

PR #34, opened 2017-01-19
Return backends in the order specified in the config file

tpcwang:
I've noticed that if you have multiple backends, then the "default" backend used by the /search/ page is not deterministic. This can be confusing to users.


nelhage:
ah, cool. Thanks

PR #35, opened 2017-01-19
Update readme to use grpc, as the frontend no longer works with the legacy method

tpcwang:
Perhaps the -listen option and relevant code should be removed from the source in a different commit?


nelhage:
ah yeah, I'd been meaning to deprecate those and forgotten. Thanks

PR #37, opened 2017-01-19
Made several frontend ui improvements

### LINKED ISSUE #14: Soft wrap long lines of code
I just got this working on a project of mine and I really like it, good work!

I did notice, however, that whenever I had a long single line of code that it wasn't being soft wrapped and instead pushes off the screen:

Would it be possible to investigate introducing a soft wrap like on some text editors:

tpcwang:
Sorry for lumping several changes together. I don't have the time to separate them again but wanted to contribute. I can understand if you don't want to take up this PR.

The changes are the following:
1. Added highlighting of the match term. This makes it easy to see what was matched.
2. Alternate the highlighting of the lines. This makes it easy to parse the context lines.
3. Added wrapping for long lines. This fixes #14 . This did complicate things as I wanted to make sure that selecting and copying the context will not add arbitrary blank lines.

I've tested this with Firefox and Chrome, so there could be issues with other browsers.


nelhage:
very nice! happy to take this, and thanks for the patches!

PR #39, opened 2017-03-17
livegrep-github-reindex: support GitHub enterprise

zarvox:
GitHub Enterprise interactions work essentially the same as those with
the GitHub public API, but with a different API base URL.  Happily,
`go-github` already supports this; we just need a flag to allow users
to point the indexer at their GitHub Enterprise base URL.

### DIFF
--- a/cmd/livegrep-github-reindex/main.go
+++ b/cmd/livegrep-github-reindex/main.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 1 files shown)


cmd/livegrep-github-reindex/main.go
```diff
@@ -81,6 +83,17 @@ func main() {
 
 	gh := github.NewClient(h)
 
+	if *flagApiBaseUrl != "" {
+		if (*flagApiBaseUrl)[len(*flagApiBaseUrl)-1:] != "/" {
```
nelhage: [`strings.HasSuffix`](https://golang.org/pkg/strings/#HasSuffix) would make this significantly less line-noise-y.

nelhage:
Thanks! I'll merge and do the one cleanup myself.

zarvox:
Thanks!  First time writing Go, sorry for the nonidiomatic usage. :)

Issue #46, opened 2017-05-16
codesearch crashes on empty repo


bnewbold:
If you run codesearch on an empty git repo (eg, one with no branches or commits yet), it crashes:

    (gdb) run repos/fail.json
    Starting program: /home/bnewbold/.cache/bazel/_bazel_bnewbold/9bebf2ad52920ae55fb82a2d3094cf32/execroot/livegrep/bazel-out/local-fastbuild/bin/src/tools/codesearch repos/fail.json
    [Thread debugging using libthread_db enabled]
    Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
    [New Thread 0x7ffff6f80700 (LWP 11764)]
    [New Thread 0x7ffff677f700 (LWP 11765)]
    [New Thread 0x7ffff5f7e700 (LWP 11766)]
    [New Thread 0x7ffff577d700 (LWP 11767)]
    Walking repo_spec name=internetarchive/community_tools, path=repos/internetarchive/community_tools
    walking HEAD... codesearch: ./src/smart_git.h:69: smart_object<T>::operator T*() [with T = git_commit]: Assertion `obj_' failed.

    Thread 1 "codesearch" received signal SIGABRT, Aborted.
    __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
    51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
    (gdb) bt
    #0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
    #1  0x00007ffff6fb53fa in __GI_abort () at abort.c:89
    #2  0x00007ffff6face37 in __assert_fail_base (fmt=<optimized out>, assertion=assertion@entry=0x555555a17d05 "obj_",
        file=file@entry=0x555555a17ce9 "./src/smart_git.h", line=line@entry=69,
        function=function@entry=0x555555a17e80 <smart_object<git_commit>::operator git_commit*()::__PRETTY_FUNCTION__> "smart_object<T>::operator T*() [with T = git_commit]") at assert.c:92
    #3  0x00007ffff6facee2 in __GI___assert_fail (assertion=0x555555a17d05 "obj_",
        file=0x555555a17ce9 "./src/smart_git.h", line=69,
[...]

PR #47, opened 2017-05-16
indexer: check whether ref resolved to commit

bnewbold:
This fixes ("works for me") issue #46.

No regression test is included. I started writing a higher level test that would call `build_index()` or `git_indexer`'s `walk()` against `doc/example/index.json` (and other example json files), but got bogged down in not knowing how to bazel.

Thank you for maintaining livegrep!

### DIFF
--- a/src/git_indexer.cc
+++ b/src/git_indexer.cc
@@ -43,7 +43,10 @@ git_indexer::~git_indexer() {
 void git_indexer::walk(const string& ref) {
     smart_object<git_commit> commit;
     smart_object<git_tree> tree;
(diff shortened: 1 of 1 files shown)


nelhage:
Thanks! I tweaked the error here a bit, but merged. 2b38b625915ea77136cb73cf707fdef348917f83

(On further analysis, not sure printing the path was necessary, since it's clear from context; But I like self-contained errors where possible. Also this condition could also just mean a typoed or missing ref, so I decided to change that text)

We don't have much of a test harness for the git indexer right now, so no worries about the lack of testing.

PR #50, opened 2017-05-24
Add livegrep-fetch-reindex helper command

jboning:
Inspired by `livegrep-github-reindex`, this clones/fetches a number of
git repositories in parallel and then regenerates the index. Repository
information is provided in the index json config format used by
`codesearch`, with a `remote` field in the `metadata` dictionary
indicating where to fetch the repository from.

The code is almost entirely plagiarized from `livegrep-github-reindex`.
`livegrep-github-index` itself could probably be refactored to generate
an index config with `remote` fields and then invoke this.

### DIFF
--- a/cmd/livegrep-fetch-reindex/main.go
+++ b/cmd/livegrep-fetch-reindex/main.go
@@ -1,0 +1,185 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+	"sync"
+)
+
+type IndexConfig struct {
+	Name         string       `json:"name"`
+	Repositories []RepoConfig `json:"repositories"`
+}
+
+type RepoConfig struct {
+	Path      string            `json:"path"`
+	Name      string            `json:"name"`
+	Revisions []string          `json:"revisions"`
+	Metadata  map[string]string `json:"metadata"`
+}
+
+var (
+	flagIndexPath   = flag.String("out", "livegrep.idx", "Path to write the index")
+	flagRevparse    = flag.Bool("revparse", true, "whether to `git rev-parse` the provided revision in generated links")
+	flagSkipMissing = flag.Bool("skip-missing", false, "skip repositories where the specified revision is missing")
+)
+
+const Workers = 8
+
+func main() {
+	flag.Parse()
+	log.SetFlags(0)
+
+	if len(flag.Args()) != 1 {
+		log.Fatal("Expected exactly one argument (the index json configuration)")
+	}
+
+	data, err := ioutil.ReadFile(flag.Arg(0))
+	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/livegrep-fetch-reindex/main.go
```diff
@@ -0,0 +1,185 @@
...
+		tmp,
+		"--index_only",
+	}
+	if *flagRevparse {
+		args = append(args, "--revparse")
+	}
+	args = append(args, flag.Arg(0))
+
+	cmd := exec.Command(
+		path.Join(path.Dir(path.Dir(path.Dir(os.Args[0]))), "src", "tools", "codesearch"),
```
nelhage: Hmm. My deployments all use the tarballs generated by `package.sh`, which copies all the binaries into a shared directory (and so does `livegrep-github-reindex`). It'd be nice to keep these in sync. It's possible I should rework my build toolchains to just export from `bazel-bin`. I'll let this lie for now and give it some thought.

cmd/livegrep-fetch-reindex/main.go
```diff
@@ -0,0 +1,185 @@
...
+			return err
+		}
+		if err := os.MkdirAll(r.Path, 0755); err != nil {
+			return err
+		}
+		return retryCommand("git", []string{"clone", "--mirror", remote, r.Path})
+	}
+
+	if err := exec.Command("git", "-C", r.Path, "remote", "set-url", "origin", remote).Run(); err != nil {
+		return err;
```
nelhage: Please `gofmt` your go code. I'll add a test to the livegrep build when I have a chance.

cmd/livegrep-fetch-reindex/main.go
nelhage: I'd also probably `path.Join` with a bunch of `..`s instead of layering `path.Dir`; it's more readable IMO.

jboning: Aha. My deployments are similarly structured, but for development I was running straight out of `bazel-bin`. (So this won't work when I deploy it, come to think of it.)

One pattern I've seen is to declare the other binary as a data dependency and call it out of the runfiles directory. I included the `data` declaration, but didn't use it.

cmd/livegrep-fetch-reindex/main.go
jboning: Hm. I thought my editor was taking care of that for me, since I saw it fixing up indentation in ways contrary to my usual configuration. Will investigate.

cmd/livegrep-fetch-reindex/main.go
nelhage: Yeah, I haven't been using `bazel run` because then (e.g.) the sandbox prevents me from accessing external git repositories, etc. And so I haven't bothered to use the runfiles directory at all.

I think I'll default both of them to 'same directory' but add a `-codesearch` flag to give an out for debugging or alternate deployments.

PR #51, opened 2017-05-24
Group multiple search results within a file

jboning:
This was originally authored by @christoffer on an old fork of livegrep; I've taken his code, dropped it onto the modern livegrep mainline, and lightly tweaked a few things.

This is a significant change to the search result UI, so I recommend playing with it a bit to see how you like it. I think the most notable change in terms of user interaction is that the click-through link is the result snippet itself, rather than the header with repo/filename, so it's now impossible to copy text directly from the result snippet.

screenshot 2017-05-23 21 57 10

### DIFF
--- a/web/htdocs/assets/js/codesearch_ui.js
+++ b/web/htdocs/assets/js/codesearch_ui.js
@@ -59,7 +59,9 @@ var MatchView = Backbone.View.extend({
   },
   _renderLno: function(n, isMatch) {
     var lnoStr = n.toString() + (isMatch ? ":" : "-");
-    return h.a({cls: 'lno-link', href: this.model.url(n)}, [
+    var classes = ['lno-link'];
+    if (isMatch) classes.push('matchline');
+    return h.a({cls: classes.join(' '), href: this.model.url(n)}, [
       h.span({cls: 'lno', 'aria-label': lnoStr}, []),
     ]);
   },
@@ -72,17 +74,17 @@ var MatchView = Backbone.View.extend({
 
     var lines_to_display_before = Math.max(0, ctxBefore.length - (clip_before || 0));
     for (i = 0; i < lines_to_display_before; i ++) {
-      ctx_before.unshift(h.div({cls: 'line'}, [
-                                 this._renderLno(lno - i - 1, false),
-                                 h.span([this.model.get('context_before')[i]]),
-                               ]));
+      ctx_before.unshift(
+        this._renderLno(lno - i - 1, false),
+        h.span([this.model.get('context_before')[i]]),
+      );
     }
     var lines_to_display_after = Math.max(0, ctxAfter.length - (clip_after || 0));
     for (i = 0; i < lines_to_display_after; i ++) {
-      ctx_after.push(h.div({cls: 'line'}, [
-                             this._renderLno(lno + i + 1, false),
-                             h.span([this.model.get('context_after')[i]]),
-                           ]));
+      ctx_after.push(
+        this._renderLno(lno + i + 1, false),
+        h.span([this.model.get('context_after')[i]]),
+      );
     }
     var line = this.model.get('line');
     var bounds = this.model.get('bounds');
@@ -95,13 +97,15 @@ var MatchView = Backbone.View.extend({
     if(clip_after !== undefined) classes.push('clip-after');
 
     var matchElement = h.div({cls: classes.join(' ')}, [
-        h.div({cls: 'contents'}, [
-                ctx_before,
-                h.div({cls: 'line matchline'}, [
-                  this._renderLno(lno, true),
-                  h.span([pieces[0], h.span({cls: 'matchstr'}, [pieces[1]]), pieces[2]]),
-                ]),
-                ctx_after])]);
+      h.div({cls: 'contents'}, [].concat(
+        ctx_before,
+        [
+            this._renderLno(lno, true),
+            h.span({cls: 'matchline'}, [pieces[0], h.span({cls: 'matchstr'}, [pieces[1]]), pieces[2]]),
+        ],
+        ctx_after,
+      )),
+    ]);
 
     return matchElement;
   }
--- a/web/htdocs/assets/js/codesearch_ui.js
+++ b/web/htdocs/assets/js/codesearch_ui.js
@@ -62,7 +62,7 @@ var MatchView = Backbone.View.extend({
     var classes = ['lno'];
     if (isMatch) classes.push('matchlno');
     return h.a({cls: 'lno-link', href: this.model.url(n)}, [
-      h.span({cls: classes.join(' '), 'aria-label': lnoStr}, []),
+      h.span({cls: classes.join(' '), 'aria-label': lnoStr}, [])
     ]);
   },
   _render: function() {
@@ -76,14 +76,14 @@ var MatchView = Backbone.View.extend({
     for (i = 0; i < lines_to_display_before; i ++) {
       ctx_before.unshift(
         this._renderLno(lno - i - 1, false),
-        h.span([this.model.get('context_before')[i]]),
+        h.span([this.model.get('context_before')[i]])
       );
     }
     var lines_to_display_after = Math.max(0, ctxAfter.length - (clip_after || 0));
     for (i = 0; i < lines_to_display_after; i ++) {
       ctx_after.push(
         this._renderLno(lno + i + 1, false),
-        h.span([this.model.get('context_after')[i]]),
+        h.span([this.model.get('context_after')[i]])
       );
     }
     var line = this.model.get('line');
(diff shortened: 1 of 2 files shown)


jboning:
There's some bug lurking in the result-combining logic which causes lines sometimes to be duplicated (such as line 2 in [this example](https://cloud.githubusercontent.com/assets/376107/26387635/fc66f49e-4003-11e7-9da0-9a3049082b47.png)). If you'd like I can track it down and update this PR with a fix.

tpcwang:
This is pretty neat! Losing the ability to copy text from the snippet is a bummer for me though. Occasionally, I like to just grep for a piece of information that I know is available elsewhere, like the version of a common dependency, for example. Being able to just copy from the search result saves me from a click.

web/htdocs/assets/js/codesearch_ui.js
```diff
@@ -72,76 +147,78 @@ var Match = Backbone.Model.extend({
...
+    });
+
+    for(var i = 1, len = this.matches.length; i < len; i++) {
+      var previous_match = this.matches[i - 1], this_match = this.matches[i];
+      var last_line_of_prev_context = previous_match.get('lno') + previous_match.get('context_after').length;
+      var first_line_of_this_context = this_match.get('lno') - this_match.get('context_before').length;
+      var num_intersecting_lines = (last_line_of_prev_context - first_line_of_this_context) + 1;
+      if(num_intersecting_lines >= 0) {
+        // The matches are intersecting or share a boundary.
+        // Split the context between the previous match and this one. Uneven splits should leave
```
tpcwang: I think this is the culprit for the bug that you are seeing. This splitting algorithm doesn't work all the time if the two matches do not have a balanced number of context lines. This typically occurs at the beginning or end of the file. For example:

previous_match = {'lno': 1, 'context_before':[], 'context_after':[2, 3, 4]}
this_match = {'lno': 2, 'context_before':[1], 'context_after': [3, 4, 5]}

The algorithm will try to split it even and remove 2 lines from previous_match and 2 lines from this_match. Resulting in:
previous_match = {'lno': 1, 'context_before':[], 'context_after':[2]}
this_match = {'lno': 2, 'context_before':[], 'context_after': [3, 4, 5]}

nelhage:
Yeah, linkifying the whole text is a bit of a bummer. What about making the header link to the file (maybe to the first hit, so that in single-match cases it behaves the same), and perhaps making the line numbers clickable links to specific lines?

Also, thanks @tpcwang for tracking down that splitting bug! I'd love to get that fixed before we land this.

web/htdocs/assets/js/codesearch_ui.js
gnprice: :trumpet: :tada: Good catch, thanks!

I reviewed this code when we originally added it to our internal fork at Dropbox in 2015, and missed this issue then, so glad to see it resolved.

web/htdocs/assets/js/codesearch_ui.js
```diff
@@ -62,86 +138,101 @@ var Match = Backbone.Model.extend({
...
+      var last_line_of_prev_context = previous_match.get('lno') + previous_match.get('context_after').length;
+      var first_line_of_this_context = this_match.get('lno') - this_match.get('context_before').length;
+      var num_intersecting_lines = (last_line_of_prev_context - first_line_of_this_context) + 1;
+      if(num_intersecting_lines >= 0) {
+        // The matches are intersecting or share a boundary.
+        // Try to split the context between the previous match and this one.
+        // Uneven splits should leave the latter element with the larger piece.
+
+        // split_at will be the first line number grouped with the latter element.
+        var split_at = parseInt(Math.ceil((previous_match.get('lno') + this_match.get('lno')) / 2.0), 10);
```
tpcwang: Given your change to linkify just the line numbers, I think there's no longer a need to split this evenly. Perhaps this chunk can be simplified to something like this?

        var clip_for_previous = Math.min(num_intersecting_lines, previous_match.get('context_after').length);
        var clip_for_this = num_intersecting_lines - clip_for_previous;

web/htdocs/assets/js/codesearch_ui.js
```diff
@@ -45,10 +47,84 @@ function shorten(ref) {
...
+    var bounds = this.model.get('bounds');
+    var pieces = [line.substring(0, bounds[0]),
+                  line.substring(bounds[0], bounds[1]),
+                  line.substring(bounds[1])];
+
+    var classes = ['match'];
+    if(clip_before !== undefined) classes.push('clip-before');
+    if(clip_after !== undefined) classes.push('clip-after');
+
+    var matchElement = h.div({cls: classes.join(' ')}, [
```
tpcwang: For what it's worth, I continued to use tables previously because I found out that chrome doesn't handle user-select: none correctly, so I think this will end up copying the line numbers.

Another benefit that table has is that long lines will not overflow into the line number column. Granted, given that the row is no longer colored, overflowing into the line number column may actually be less confusing.

image

image

tpcwang: By the way, my comment above is just a note for reference, I like the previous behavior of not copying the lno, but it might not be all that important?

jboning: I agree that not copying the lno is preferable. (Though perhaps sometimes you might want to copy just the lno, to jump to that line in a text editor or something? Probably not a major use case.)

I was fooled by the fact that Chrome didn't highlight the lnos when selecting code. But sure enough, they're included in the copied text.

I'll fiddle with this some more and see if I come up with anything. Maybe returning to the table approach is the thing to do.

jboning: CSS `content:` seems to do the right thing here, though I'm sure there's some other hidden drawback I haven't thought of. That of course doesn't address the wrapping thing.

jboning: I came up with a solution for line wrapping based on CSS grid layout that seems to work OK. Let me know what you think!

web/htdocs/assets/js/codesearch_ui.js
jboning: Yeah, this can probably be simplified now. But we just fixed the logic! 😑

web/htdocs/assets/js/codesearch_ui.js
tpcwang: I think I didn't do that previously because it doesn't work on IE, but I can live with IE not working :)

Thanks for fixing this!

web/htdocs/assets/js/codesearch_ui.js
tpcwang: Yeah, your logic is correct, I just pointed it out since it was a little bit complicated to reason through.

web/htdocs/assets/js/codesearch_ui.js
```diff
@@ -45,10 +47,89 @@ function shorten(ref) {
...
+    var ctx_before = [], ctx_after = [];
+    var lno = this.model.get('lno');
+    var ctxBefore = this.model.get('context_before'), clip_before = this.model.get('clip_before');
+    var ctxAfter = this.model.get('context_after'), clip_after = this.model.get('clip_after');
+
+    var lines_to_display_before = Math.max(0, ctxBefore.length - (clip_before || 0));
+    for (i = 0; i < lines_to_display_before; i ++) {
+      ctx_before.unshift(
+        this._renderLno(lno - i - 1, false),
+        h.span([this.model.get('context_before')[i]]),
```
tpcwang: IE doesn't like the extra ',' here and a few other places. Even with this fixed, IE still won't display these correctly, but at least it will load the js without error.

web/htdocs/assets/js/codesearch_ui.js
jboning: Yikes. I can't even figure out how to make IE's `-ms-grid` layout do the right thing here. I'm not sure whether to abandon IE/Edge, or switch back to a table-based layout, where pasting as formatted text is weird.

Github uses a table, here's what happens when pasting into Gmail on Chrome:
screenshot 2017-05-26 14 51 38

Perhaps I should give up on the idea of doing better, since Github are professionals and this is the best they've got (and I'm clearly not a web expert).

gnprice: Yikes.

My suggestion would be to merge the best version you have without worrying about IE, and file any known issues with it in the tracker. Then if someone comes along who has a better idea of how to make this work uniformly, they'll have all the information to take a crack at it.

nelhage: Yeah, I'm fine to land without perfect IE support. Please fix up the trailing commas above, though.

Probably there's a linter I could/should import somewhere that would let us run CI for minor issues like that.

I've been vaguely contemplating moving to a more modern frontend stack (React?) which would hopefully also make it easier to do frontend CI without a full headless browser stack.

gnprice:
> Probably there's a linter I could/should import somewhere that would let
us run CI for minor issues like that.

I took a look at what Zulip does, since that project is pretty serious
about its linters; looks like it uses ESLint, and configures that rule like
so:
https://github.com/zulip/zulip/blob/cb0527397/.eslintrc.json#L198

Seems like a good idea.

jboning:
I've fixed up the trailing commas, so it works in IE, though it still renders very poorly.

tpcwang:
:+1: Woot

PR #72, opened 2017-06-24
Decouple file/repo/tag case sensitivity from line case sensitivity

### LINKED ISSUE #71: File/Repo case sensitivity is coupled with line case sensitivity
This is not obvious to the user, and it can be confusing.

For example, this returns no results:
https://livegrep.com/search/linux?q=Linux+project+file%3Acredits

But this does:
https://livegrep.com/search/linux?q=Linux+project+file%3ACREDITS

tpcwang:
Fixes #71

### DIFF
--- a/src/tools/grpc_server.cc
+++ b/src/tools/grpc_server.cc
@@ -10,10 +10,11 @@
 
 #include <json-c/json.h>
 
-#include <string>
 #include <algorithm>
+#include <cctype>
 #include <functional>
 #include <future>
+#include <string>
 
 #include <boost/bind.hpp>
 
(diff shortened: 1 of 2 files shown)


nelhage:
I think my intent was that `file` and `repo` should always be case-sensitive, since it seems less common to need case-folding there; Alterately, perhaps "always fold" would work. But this makes it more consistent, so I'm fine with this and we can change if needed.

tpcwang:
Agreed. I'd be fine with always case-sensitive, always case-fold, or this fix because they are all consistent. I chose this route because it seems to be more inline with what a user might expect.

Issue #65, opened 2017-05-27
Rank results by relevance


gnprice:
Once #51 is in, we'll have the power to conveniently rank some files above others in the results based on whatever we think will best help people find what they want.

#64 contains one idea for that (rank definitions, heuristically identified, highly). That one will be great but requires some infrastructure to make it work.

Here are a couple of other ranking features that would be easy to build:
* Rank files that sound like test files to the bottom. A fairly simple filename pattern (just a little subtler than `/test/i`) should do a good job here.
* Maybe rank files with lots of hits to the top. I'm not 100% sold on this as a good feature, but I've heard the suggestion and it seems plausible.

Bonus feature: provide the same is-this-a-test pattern as an option in queries, e.g. spelled `is:test` (or as people will probably use it more often, `-is:test`.)

PR #75, opened 2017-06-30
Integrated file search

### LINKED ISSUE #17: Search filenames
Doable now with `. file:FILENAME$`, but it'd be nice if it didn't require the `.` trick (and thus matching on every line of the file).

(I know, I know, just send in a pr. Consider this a reminder to myself to work on it when I have time, unless someone else feels like taking a stab at it first.)

jboning:
Run the input query against file paths as well as file contents, and show matches before the main results.

The filename index isn't stored in the on-disk format, but only held in memory, and rebuilt every time the index is loaded. On our corpus this adds about 2.5s of startup time.

I'm not totally happy with the code quality (copied JS; crudely refactored suffix_search function) or the UX (probably want a shorter list of filename results and some heuristics on whether they're interesting to display), but it's a start!

Fixes #17.

### DIFF
--- a/src/codesearch.cc
+++ b/src/codesearch.cc
@@ -207,10 +207,10 @@ class searcher {
 public:
     searcher(const code_searcher *cc,
              const query &q,
-             const intrusive_ptr<IndexKey> index,
+             const intrusive_ptr<IndexKey> index_key,
              const code_searcher::search_thread::transform_func& func) :
         cc_(cc), query_(&q), transform_(func), queue_(),
-        matches_(0), limiter_(q.max_matches), index_(index), re2_time_(false),
+        matches_(0), limiter_(q.max_matches), index_key_(index_key), re2_time_(false),
         git_time_(false), index_time_(false), sort_time_(false),
         analyze_time_(false), files_(new uint8_t[cc->files_.size()]),
         files_density_(-1)
@@ -344,7 +344,7 @@ protected:
     thread_queue<match_result*> queue_;
     atomic_int matches_;
     search_limiter limiter_;
-    intrusive_ptr<IndexKey> index_;
+    intrusive_ptr<IndexKey> index_key_;
     timer re2_time_;
     timer git_time_;
     timer index_time_;
@@ -367,8 +367,9 @@ class filename_searcher {
 public:
     filename_searcher(const code_searcher *cc,
                       const query &q,
-                      intrusive_ptr<IndexKey> index) :
-        cc_(cc), query_(&q), index_(index), queue_(), matches_(0), limiter_(q.max_matches) {}
+                      intrusive_ptr<IndexKey> index_key) :
+        cc_(cc), query_(&q), index_key_(index_key), queue_(), matches_(0), limiter_(q.max_matches)
+    {}
 
     void operator()();
 
@@ -377,7 +378,7 @@ protected:
 
     const code_searcher *cc_;
     const query *query_;
-    intrusive_ptr<IndexKey> index_;
+    intrusive_ptr<IndexKey> index_key_;
     thread_queue<file_result*> queue_;
     atomic_int matches_;
     search_limiter limiter_;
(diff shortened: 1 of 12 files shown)


src/codesearch.cc
```diff
@@ -352,6 +368,49 @@ class searcher {
...
+    thread_queue<file_result*> queue_;
+    atomic_int matches_;
+    search_limiter limiter_;
+
+    friend class code_searcher::search_thread;
+};
+
+void filename_searcher::operator()(indexed_file *file)
+{
+    if (limiter_.why())
```
nelhage: Hmm. I wonder if we should rename `why()`; It is intended only to be called post-facto, and `exit_early()` should be called in the loop. I think I short-circuited to checking `exit_reason_` either over concern about the  `gettimeofday` overhead (likely overblow; gtod is super fast on modern systems), or out of some kind of consistency concern (but this is multi-threaded, so that seems unlikely).

Possibly we should just do the full check on every query, and if that's too much we can move the `CODESEARCH_SLOWGTOD` hack into the default path with a smaller constant…

src/codesearch.cc
```diff
@@ -439,10 +444,24 @@ code_searcher::~code_searcher() {
     }
 }
 
+void code_searcher::chunk_files() {
```
nelhage: `chunk_files` is a super confusing name for this, since we also have a data structure called `chunk_file` :)

src/codesearch.cc
```diff
@@ -206,18 +207,15 @@ class searcher {
 public:
     searcher(const code_searcher *cc,
              const query &q,
+             const intrusive_ptr<IndexKey> index,
```
nelhage: can we call this `key` or something instead of `index`? "index" refers in my mind to the data structure with the suffix arrays, and an "index key" is a thing that one can look up efficiently in it.

src/codesearch.cc
jboning: That's helpful context. I'll adjust here and maybe elsewhere too.

src/codesearch.cc
jboning: Heh! This went away in the later commit where I started using a proper index (renamed to `index_filenames` or something), so I'll see about rebasing this away.

src/codesearch.cc
jboning: Sure! I opted for consistency despite being a bit puzzled by the naming, so I'm happy to make that change :)

jboning:
Yeah, you have the idea exactly. Anecdotally (based on experience with a different code search tool) I've found it quite handy to be able to find a file I want to look at quickly just by typing (part of) the filename.

Before landing, I think I want to add a bit of UI to limit the number of filename hits displayed (perhaps 5 or 10, with a button to expand to more). Eventually we'll probably want some smarter heuristics to decide whether/which filename hits are displayed (related to #65).

Here's an example where it works pretty well:

screenshot 2017-07-05 23 04 58

Here's an example where the code as it stands produces a tragedy and we should clearly show fewer filename results:

screenshot 2017-07-05 23 06 55

src/codesearch.cc
jboning: Looking at the code here:
 * This particular use is no longer inside a loop after adding an index on filenames. I don't think I'm going to bother to expunge it from the history. (Unless you'd prefer to squash the commits so that there's one commit that changes the RPC interface and sets up the index all at once.)
 * Another potential source of slowness when using `exit_early()` is the `matches_.load()`. I don't know how slow an `atomic_int::load()` is, but it seems like it might be a bit expensive. So I'll go ahead and add another version of `exit_early()` that doesn't include the match-limit check, for use when we know the match limit hasn't changed.

jboning: OK, I ended up refactoring the matches counter to live inside the limiter, and gave the limiter a `record_match()` method instead, which feels much nicer.

We could possibly accept the possibility of a data race and use a nonatomic int here, because:
 * this count is only used for the search limiter, so it's acceptable to be imprecise
 * limiting is already a bit imprecise--two threads can push results onto the queue in `try_match()` when strict limit enforcement would mean only one of them should
 * if correctness is a concern, we can apply limit separately when reading results from the queue later
But, there's potential for the count to much more wildly diverge from reality, and I think probably the performance win isn't significant, so I'm not going to make that change.

jboning:
Whoops, forgot that github doesn't notify about updates to PRs. I believe I've addressed the comments so far, so this should be ready for another look.

PR #76, opened 2017-07-05
Add rudimentary stats reporting

jboning:
- Create a /debug/stats route
 - Include index creation time in Info RPC result
 - Report index age as a stat

It's not clear whether this is the right long-term approach for stats (maybe we'd rather speak statsd or something?).

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -161,6 +161,29 @@ func (s *server) ServeHealthcheck(w http.ResponseWriter, r *http.Request) {
 	io.WriteString(w, "ok\n")
 }
 
+type stats struct {
+	IndexAge int64 `json:"index_age"`
+}
+
+func (s *server) ServeStats(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+	// For index age, report the age of the stalest backend's index.
+	now := time.Now()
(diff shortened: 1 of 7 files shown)


alexmv:
I like exposing the stats as JSON, because likely every site is going to have their own pipieline for ingesting that into a central data store.

nelhage:
Yeah, I'm a fan of exporting stats as JSON over speaking `statsd` ourselves.

server/server.go
```diff
@@ -161,6 +161,29 @@ func (s *server) ServeHealthcheck(w http.ResponseWriter, r *http.Request) {
...
+type stats struct {
+	IndexAge int64 `json:"index_age"`
+}
+
+func (s *server) ServeStats(ctx context.Context, w http.ResponseWriter, r *http.Request) {
+	// For index age, report the age of the stalest backend's index.
+	now := time.Now()
+	maxBkAge := time.Duration(-1) * time.Second
+	for _, bk := range s.bk {
+		if bk.I.IndexTime == time.Unix(0, 0) {
```
nelhage: `time.Time` has an `IsZero` you could use.

jboning: Aha, thanks for the tip!

PR #79, opened 2017-08-24
Rework query UI

### LINKED ISSUE #58: Literal search by default, regex optional
This is a change we made in our internal livegrep instance at Dropbox, and I've been quite happy with it.

In my experience it's been much more common that I'm searching for a literal string that involves a paren or bracket or backslash or something and needs to be escaped than that I actually want to search for a regex. Regex search is irreplaceable when it's needed, but flipping the default has made the experience smoother.

The way I'd imagine doing this is with a `re:` colon-operator, the antonym of `lit:`. Then search clauses with no colon-operator become implicitly `lit:` instead of `re:`. (Our internal instance is a fork from 2014, predating colon-operators, so we do it with a checkbox labeled "regex" next to the search box.)

Maybe this default should be instance-configurable -- the fast regex search is a sweet demo, so maybe `livegrep.com` should keep it front and center by keeping regex as the default.

### LINKED ISSUE #74: Query modifier UI
Inspired by #71, filing a separate issue to discuss query modifier UI in general.

The query parser currently understands these tags:
 - `file:`
 - `repo:`
 - `tags:`
 - `case:`
 - `lit:`
 - `max_matches:`

I'd group these as follows:
 - The `file:` and `repo:` tags are like extra search terms, filtering query results. (`tags:` seems also to fall into this category, but additionally slightly modifies treatment of the main content query?)
 - `case:` and `lit:` are modifiers for the meaning of the main query.
 - `max_matches:` is a modifier on the overall search behaviour.

For the second group (`case` and `lit`), there are a few rough edges:
 - The default behaviour/setting is not obvious (though it's documented on the help page).
 - If a user typically uses a particular mode, the setting isn't sticky.
 - It's not clear how these modifiers interact with `file` and `repo` terms--whether you can ask for `file:lit:foo.c`. Since you can't, there's no way to control the case-sensitivity of a `file` field (after #72).

---

One possibility is to move the query modifiers into separate UI controls, something like this:

    Case:  (x) auto   ( ) match   ( ) ignore
    Query: (x) regex  ( ) literal

This:
 - makes the query interpretation explicit in the UI (especially the "auto" case-sensitivity detection, which I didn't even realize existed!)
 - lets us store the user's case/query preferences in a cookie and apply them again at next page load

With this UI, I think it would be least surprising to apply the settings to both the main query and `file:` terms, which would be a partial reversal of #72.

jboning:
This is a fairly substantial overhaul of the query input UI.

The major pieces are:
 - Lift case and regex settings out into standalone UI elements
 - Place help text prominently on the main search page, when no search has been performed.
 - Add a brief hint about the special search controls directly adjacent the main query input
 - Embiggen query input

The result looks like this:
screenshot 2017-08-24 16 41 03

Fixes #74, fixes #58


tpcwang:
I didn't look at the code but making the special terms obvious and up-front seems like a good idea. I like it!

nelhage:
Nice, I like it.

PR #80, opened 2017-08-25
Link filename matches to built-in code browser when enabled

jboning:
This matches the behaviour for file-content results.

### DIFF
--- a/web/htdocs/assets/js/codesearch_ui.js
+++ b/web/htdocs/assets/js/codesearch_ui.js
@@ -47,6 +47,45 @@ function shorten(ref) {
   return ref;
 }
 
+function url(tree, version, path, lno) {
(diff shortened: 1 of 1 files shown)


web/htdocs/assets/js/codesearch_ui.js
```diff
@@ -272,6 +272,16 @@ var FileMatch = Backbone.Model.extend({
   },
 
   url: function() {
```
alexmv: The code duplication with Match, above, is unfortunate.

Any way to combine those nicely?

jboning: OK, I've pulled that logic out into some shared functions. (If I'd done that the first time, I probably could have avoided this bug in the first place!)

PR #82, opened 2017-08-25
If custom header content is provided, replace the default header

jboning:
Since the tech-demo-style header from livegrep.com don't necessarily make sense for a corporate internal instance.

### DIFF
--- a/web/htdocs/assets/css/codesearch.css
+++ b/web/htdocs/assets/css/codesearch.css
@@ -162,6 +162,7 @@ a:hover {
     width: 60em;
     margin: auto;
     color: #999;
+    padding-bottom: 100px;
 }
 
 .helpsection {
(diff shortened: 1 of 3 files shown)


jboning:
I don't feel great that this removes all attribution. Maybe the thing to do is wait for #74 to land, and then add some credits on the blank search page below the help text?

nelhage:
I'm happy landing this if it's useful to you, but I think a credit or `/about` or other link at the bottom of the help page in #79 would make some sense.

jboning:
I played with this briefly and ended up putting in a footer that's present below the help text or any search results. I didn't linkify it--I think this should help avoid routing internal feedback to you, while still attributing credit and making it possible for curious users to find you (using the magic of google).

PR #83, opened 2017-08-25
Fix case where 'displaying' search ID can decrease

### LINKED ISSUE #38: Uncaught TypeError: Cannot read property 'q' of undefined
Very cool project. While using search I encountered this error in the JS console:

```javascript
Uncaught TypeError: Cannot read property 'q' of undefined
```
related to this line:
```javascript
   if (this.model.search_map[this.model.get('displaying')].q === '' ||
       this.model.get('error')) {
      this.$el.hide();
      return this;
    }
```

Looks like you may only want to consider calling `.q` when `search_map` is not `undefined`. Or, you may want to change how you return data to the client.

jboning:
Duplicates check from the other result handle_* functions.

Fixes #38

### DIFF
--- a/web/htdocs/assets/js/codesearch_ui.js
+++ b/web/htdocs/assets/js/codesearch_ui.js
@@ -428,6 +428,8 @@ var SearchState = Backbone.Model.extend({
     }
   },
   handle_done: function (search, time, why) {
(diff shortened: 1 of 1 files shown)


nelhage:
Nice, thanks.

PR #85, opened 2017-08-29
Restore hidden backend select input

### LINKED ISSUE #84: All livegrep search result links are "/null"
When I back up to b459d1f the livegrep README instructions work fine, but when I advance to the next master commit 051b043 then the service, when run using the same README instructions, presents search rules that when moused over all offer only the link http://localhost:8910/null instead.

jboning:
I accidentally dropped this in #79. Fixes #84.

### DIFF
--- a/web/templates/index.html
+++ b/web/templates/index.html
@@ -55,6 +55,9 @@
       </div>
     {{else}}
       {{with index .Backends 0}}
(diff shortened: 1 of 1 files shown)


nelhage:
Thanks for the quick fix!

jboning:
Eh, just fixing my own mistake :)

PR #90, opened 2017-08-30
Bugfix for 'new search' link in file browser

jboning:
Beginning a new search by clicking the link shouldn't prefill the query `[object Object]`.

In #81 I either broke this or failed to fix it.


alexmv:
Also update the commit description to clarify this is about the link?  Otherwise LGTM

jboning:
Repushed with updated commit message.

PR #91, opened 2017-08-30
Hacky support for bare file: queries

### LINKED ISSUE #89: support standalone `file:` queries
A query of `file:foo` results in an error: "You must specify a regex to match". We should allow this, and only report matching files.

jboning:
I think ultimately the right path forward is to extend `codesearch` itself with some smarter behaviour. But this was too easy of a hack for me to pass up, just in case you want to take it :)

Fixes #89

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -181,10 +181,16 @@ func (s *server) ServeAPISearch(ctx context.Context, w http.ResponseWriter, r *h
 		return
 	}
 
-	if q.Line == "" {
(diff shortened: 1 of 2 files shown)


nelhage:
Hm, this seems tempting, but I think the `codesearch` changes shouldn't be too hard, and you've nerdsniped me; I'll attempt to get to those weekend, and then consider landing this if it turns into too much of a yakshave.

jboning:
obviated by #98!

Issue #57, opened 2017-05-27
Option for less context and way more results


gnprice:
When I use `grep` or the like locally, I often use `-C` but also often don't. Skipping context is good if I have a lot of results and want to take in a comprehensive view of where there are matches and/or what the matches look like.

Currently livegrep always does the equivalent of `-C3`. It'd be great to have the option of `-C0`. (Maybe also other options like `-C1` if doing so fits naturally in the UI.)

Relatedly, sometimes I want a comprehensive picture of where some pattern appears even though there are a few hundred of them. Thanks to #49, all the infrastructure is in place for the front end to offer more matches. Could be a UI widget, maybe, or perhaps just request more when it knows it's showing less context and can handle more.

gnprice:
A variation on part of this that I've heard suggested: even when there are too many results to show them, give an estimate of how many. Handy for e.g. estimating the rough scope of a refactor.

This could be a rough estimate (the search engine already internally computes some statistics that could be used for this), or could be an actual count up to some higher limit.

PR #92, opened 2017-08-31
Add UI control to hide result context

jboning:
Partially addresses #57: this yields a much denser view, without increasing the number of results.


nelhage:
Not bad. Let's run with it.

alexmv:
This seems to not be "sticky" (pushed into the query string) like the rest of the controls.

jboning:
Indeed, it's not. I can see about doing that, for consistency.

PR #95, opened 2017-08-31
Tweak max_matches behaviour

jboning:
- If `max_matches:` is provided in the query, use that as the limit,
   regardless of a limit passed on the `codesearch` command line. This
   is on the theory that, if the user provides a `max_matches` value
   greater than the configured limit, it's because they want to see more
   results. If they want to see a LOT of results and are willing to wait
   for it, that's their choice.
 - Any time the search completes early (without an exhaustive list of
   results), indicate that the count in the UI is not exact. Included in
   this change because we're making it easier to hit the time limit
   (rather than the match limit).
 - Mention `max_matches:` in the help section.


nelhage:
For context, a lot of the limits and limit code are because I was irrationally concerned about DoS when I first launched livegrep -- casualty of working in security around that time, perhaps :)

Empirically, that's not a problem I've had, and so giving more flexibility to users makes a lot of sense to me.

PR #97, opened 2017-08-31
Drop search permalink

jboning:
It's redundant with the URL.


nelhage:
:+1: I feel like tabs changing their own URL was less common when I wrote this, but it does seem redundant now.

PR #98, opened 2017-09-03
Handle filename-only queries

nelhage:
Similar to #91, but we plumb through to the backend.

### DIFF
--- a/server/query.go
+++ b/server/query.go
@@ -127,7 +127,21 @@ func ParseQuery(query string) (pb.Query, error) {
 			bits = append(bits, bit)
 		}
 	}
-	out.Line = strings.Join(bits, "")
+
+	if len(bits) > 1 {
+		return out, errors.New("You cannot provide multiple of case:, lit:, and a bare regex")
+	}
+
+	if len(bits) > 0 {
+		out.Line = bits[0]
+	}
+
+	if out.Line == "" && out.File != "" {
+		out.Line = out.File
+		out.File = ""
+		out.FilenameOnly = true
+	}
+
 	if _, ok := ops["case"]; ok {
 		out.FoldCase = false
 	} else if _, ok := ops["lit"]; ok {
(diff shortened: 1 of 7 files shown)


server/query.go
```diff
@@ -127,7 +127,21 @@ func ParseQuery(query string) (pb.Query, error) {
...
+
+	if len(bits) > 1 {
+		return out, errors.New("You cannot provide multiple of case:, lit:, and a bare regex")
+	}
+
+	if len(bits) > 0 {
+		out.Line = bits[0]
+	}
+
+	if out.Line == "" && out.File != "" {
```
jboning: Just out of curiosity: Why did you decide put this logic here, vs pushing to the backend?

Mainly I'm trying to develop my intuition for the division of labor in the design :)

src/codesearch.cc
```diff
@@ -1109,34 +1113,45 @@ void code_searcher::search_thread::match(const query &q,
...
     while (file_search.queue_.pop(&f)) {
+        file_matches++;
         fcb(f);
         delete f;
     }
 
-    search.get_stats(stats);
+    if (q.filename_only) {
+        stats->why = file_search.why();
+        stats->matches = matches;
```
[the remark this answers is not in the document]
nelhage: fixed

server/query.go
nelhage: I think my basic intuition is that I want to minimize arg-parsing or deep interpretation of the `Query` struct in `codesearch`, in place of keeping it as straightforward as possible an implementation of the RPC described in the proto, and I want `livegrep` to be the place that handles fiddly UX choices and the squishy problem of mapping between human inputs and something structured.

I'm not sure this is clearly right, since (e.g.) the best UX might mean pushing that squishiness all the way back, but that's my current approximate framework.

jboning: Cool, that makes sense.

PR #100, opened 2017-09-06
Make frontend smarter about filename-only searches

jboning:
Tell the frontend explicitly whether a search was filename-only or not;
if it was, render more results, and show the number of file results in
the UI.


jboning:
Milestone PR #100 here!

PR #101, opened 2017-09-06
Increase variation in page titles

jboning:
- Drop "code search" from the title of every page; retain it only as the
  title of the search page with no query
- Search results have title "$query - search"
- Browser has file or directory name as title
- Add a mediocre favicon to help identify livegrep tabs; this makes
  dropping "code search" less of a loss


jboning:
@nelhage I'm interested in your take on the favicon in particular, since that's something approximating branding :)

nelhage:
Seems better than nothing :) I've got no particular attachment to the visual identity, and it's an improvement :)

PR #103, opened 2017-09-07
Use dropbox/rules_node to build JS packages

jboning:
- Load jquery, backbone, and underscore from cdnjs instead of hosting
   locally (and removes them from the local repo)
 - Use webpack (via bazel) to build a JS bundle for each page

This removes all vendored third-party js libraries (except for html.js;
I couldn't find that in NPM, and anyway it's not that big).

JS source is organized under web/src/, and the road is paved to
refactor it into files/modules.

The major drawback is that now `livegrep` must be run with -docroot set
to a bazel-built location (I've been passing `-docroot
bazel-bin/cmd/livegrep/livegrep.runfiles/com_github_livegrep_livegrep/web/`),
and JS changes require a rebuild (`bazel build //web:webpack_build`
is usually good enough).


nelhage:
It'd be nice to get a development mode that serves assets directly without needing a rebuild, but I don't think it's a blocker to landing.

nelhage:
:+1: from me. Might make sense to include a wrapper script that points `livegrep` at the right thing, or make it default to that path?

jboning:
OK, I've
 * rebased to fix merge conflicts
 * switched to a single JS bundle
 * made the docroot default to runfiles

The CI error looks spooky:
```
Error downloading [https://github.com/libgit2/libgit2/archive/v0.24.1.tar.gz] to /home/travis/bazel/external/com_github_libgit2/libgit2-0.24.1.tar.gz: Checksum was 0269ec198c54e44f275f8f51e7391681a03aa45555e2ab6ce60b0757b6bde3de but wanted 60198cbb34066b9b5c1613d15c0479f6cd25f4aef42f7ec515cd1cc13a77fede
```

jboning:
It looks like the release tarball was updated to include tests: https://github.com/libgit2/libgit2/issues/4343#issuecomment-328709254

jboning:
Nope; it looks like the bazel cache just didn't include the `tests` directory. The contents are completely identical. I'll update the pin in WORKSPACE and rebase.

PR #104, opened 2017-09-10
Added file hint to index.html

tammybutow:
Small request from users during our demo. 

They said it would be good to know searching using file: was possible and recommended adding it to the hints.


tammybutow:
small change to index.html for search ux

PR #109, opened 2017-09-22
Query parser tweaks

jboning:
This makes the parsing more intuitive (to me). Some of this might be contentious.

The most fundamental change is that this exposes to the parser whether the query is a regex or not; this information is used to modify parenthesis handling.

The test case changes provide a good summary of the behaviour changes.

### DIFF
--- a/server/query.go
+++ b/server/query.go
@@ -38,11 +38,11 @@ func onlyOneSynonym(ops map[string]string, op1 string, op2 string) (string, erro
 	return ops[op2], nil
 }
 
-func ParseQuery(query string) (pb.Query, error) {
+func ParseQuery(query string, globalRegex bool) (pb.Query, error) {
 	ops := make(map[string]string)
 	key := ""
 	q := strings.TrimSpace(query)
-	inRegex := true
+	inRegex := globalRegex
 
 	for {
 		m := pieceRE.FindStringSubmatchIndex(q)
@@ -61,10 +61,10 @@ func ParseQuery(query string) (pb.Query, error) {
 				ops[key] += " "
 			} else {
 				key = ""
-				inRegex = true
+				inRegex = globalRegex
 			}
 		} else if match == "(" {
-			if !inRegex {
+			if !(inRegex) {
 				ops[key] += "("
 			} else {
 				// A parenthesis. Nothing is special until the
@@ -129,7 +129,7 @@ func ParseQuery(query string) (pb.Query, error) {
 	var bits []string
 	for _, k := range []string{"", "case", "lit"} {
 		bit := strings.TrimSpace(ops[k])
-		if k == "lit" {
+		if k == "lit" || !globalRegex {
 			bit = regexp.QuoteMeta(bit)
 		}
 		if len(bit) != 0 {
@@ -143,6 +143,13 @@ func ParseQuery(query string) (pb.Query, error) {
 
 	if len(bits) > 0 {
 		out.Line = bits[0]
+	}
+
+	if !globalRegex {
+		out.File = regexp.QuoteMeta(out.File)
+		out.NotFile = regexp.QuoteMeta(out.NotFile)
+		out.Repo = regexp.QuoteMeta(out.Repo)
+		out.NotRepo = regexp.QuoteMeta(out.NotRepo)
 	}
 
 	if out.Line == "" && out.File != "" {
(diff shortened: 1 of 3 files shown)


server/query_test.go
```diff
@@ -137,9 +137,49 @@ func TestParseQuery(t *testing.T) {
...
+		{
+			"a( file:b",
+			pb.Query{Line: `a\(`, File: "b", FoldCase: true},
+		},
+		{
+			"a (file:b",
+			pb.Query{Line: `a \(file:b`, FoldCase: true},
+		},
+		{
+			"(file:a b",
```
jboning: It's not completely clear to me what the "right" behaviour for these cases should be.

jboning:
Also, this makes the parser rather complicated, which I don't love. I wasn't able to come up with an obvious and cleaner way to do it, but maybe I did't try hard enough?

nelhage:
I'm fine with a more-complex parser if it's well-tested and behaves more intuitively. I'll review in a bit of detail though.

server/query_test.go
```diff
@@ -124,10 +116,18 @@ func TestParseQuery(t *testing.T) {
...
+		},
+		{
+			`lit:a(b file:c`,
+			pb.Query{Line: `a\(b`, File: "c", FoldCase: false},
+		},
 	}
 
 	for _, tc := range cases {
-		parsed, err := ParseQuery(tc.in)
+		parsed, err := ParseQuery(tc.in, true)
```
nelhage: Would it make sense to roll the "is regex" flag into the test table, instead of having two loops? That would certainly be my instinct.

If you wanted to be slightly more concise, we could even flip the polarity and have the flag be "isLiteral" so that you could omit it from the regex tests and it'd default to false.

server/query.go
```diff
@@ -38,74 +38,106 @@ func onlyOneSynonym(ops map[string]string, op1 string, op2 string) (string, erro
...
+	inRegex := globalRegex
+	justGotSpace := true
 
 	for {
 		m := pieceRE.FindStringSubmatchIndex(q)
 		if m == nil {
-			ops[key] += q
+			term += q
+			if _, alreadySet := ops[key]; alreadySet {
+				return out, fmt.Errorf("(a) got term twice: %s", key)
```
nelhage: Are these `(a)` and `(b)` for your debugging? Should we clean up these error message?

server/query.go
```diff
@@ -38,74 +38,106 @@ func onlyOneSynonym(ops map[string]string, op1 string, op2 string) (string, erro
...
 		} else if match[0] == '\\' {
-			ops[key] += match
+			term += match
 		} else {
 			// An operator. The key is in match group 1
 			newKey := match[m[2]-m[0] : m[3]-m[0]]
 			if key == "" && knownTags[newKey] {
+				if strings.TrimSpace(term) != "" {
+					if _, alreadySet := ops[key]; alreadySet {
+						return out, fmt.Errorf("main search term must be contiguous")
```
nelhage: Similarly here, I think we could do better on the error?

server/query_test.go
nelhage: I think the fact that it's unclear means that probably people won't generate them, and as long as we do something plausible (possibly we should even error?) it's fine.

Alternately, I think that if the "not regex" flag is checked, erring on the side of treating the entire box as a single literal search is pretty fair (which is what this seems to mostly do).

server/query_test.go
jboning: I think I tried that and instead of defaulting to false I just got an error that I wasn't providing enough fields to initialize. Maybe I'm misremembering, so I'll give it another go.

server/query_test.go
jboning: I don't think treating the entire box as a single literal search is right. It's still pretty common to want to filter by path. Maybe completely disregarding parentheses is the thing to do?

server/query_test.go
nelhage: Mm, that might require switching to explicit field names, which may not be a great tradeoff.

I'd weakly prefer spelling out `false` and `true` in each case over two loops, I think.

server/query_test.go
nelhage: tbc I wasn't suggesting a behavior change; The choices you've made seem defensible, I was just musing.

jboning: Happy to muse!

I think I may be leaning towards making parentheses completely un-special in literal mode. It seems reasonable to say that if you want to use parens to escape a `file:` or other term, you need to be in regex mode, where parens have a clear and specific meaning. I think that will make the query parser a bit simpler, too.

PR #110, opened 2017-09-22
Fetch static resources with content hashes

jboning:
This ensures that users don't see old cached versions of the javascript.

Enforcing subresource integrity here is probably unneeded, but we have
the information, so why not?

### DIFF
--- a/server/BUILD
+++ b/server/BUILD
@@ -31,6 +31,7 @@ go_library(
         '//web:htdocs',
         '//web:templates',
         '//web:webpack_build',
+        '//web:asset_hashes',
     ],
 )
 
--- a/server/server.go
+++ b/server/server.go
@@ -29,20 +29,27 @@ type Templates struct {
 }
 
 type server struct {
-	config  *config.Config
-	bk      map[string]*Backend
-	bkOrder []string
-	repos   map[string]config.RepoConfig
-	inner   http.Handler
-	T       Templates
-	Layout  *template.Template
+	config      *config.Config
+	bk          map[string]*Backend
+	bkOrder     []string
+	repos       map[string]config.RepoConfig
+	inner       http.Handler
+	T           Templates
+	AssetHashes map[string]string
+	Layout      *template.Template
 
 	honey *libhoney.Builder
 }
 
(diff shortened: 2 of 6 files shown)


nelhage:
Since we're only adding the asset hash to the query string, and not using hashed paths, does this mean that during a deploy we risk the server returning the other version's assets, and then failing a SRI check?

jboning:
That's a possibility, yeah. I think it's probably better than running an unexpected version of JS and getting some weird behaviour later, but I'm not confident about that.

We could replace the stock `/assets/` request handling with a custom version that ensures we're returning the requested version. That sounds clunky though.

Is there any way we can convince the bazel to produce files with names that vary based on the contents? I think that's simply ruled out by the design, but maybe there's some trick I don't know.

nelhage:
bazel needs to know the filenames it's producing ahead of time; If we wanted stamped assets, we'd need to produce a zip/tarball and/or use something like `go-bindata` to bake them into the binary. The latter wouldn't be the worst idea, but it's also nice to be able to serve them separately from a CDN or just an nginx.

Assets with hashes in their URL have the nice property that upgrades can be done in a race-free way by adding the new assets to the asset server, and then kicking the app; During the deploy, it doesn't matter if you see the pre- or post-deploy HTML, because either way you get the corresponding stamped assets.

Now, livegrep is low enough traffic and non-critical, so we probably don't particularly need to do that dance if it's hard; And I think, on reflection, SRI has the nice property that we don't risk the browser caching an asset under the wrong hash, and getting stuck again. So I'm happy to take this. 

That says, surely the node ecosystem has some builder that can also build our assets and generated stamped URLs and a stamped asset tarball, that we can eventually Just Use if it becomes a problem?

jboning:
Surely the node ecosystem has something like that. I'm never confident in being able to "just use" anything from the node ecosystem, though :)

It should be easy enough to use a bazel genrule to build an (unstamped) tarball containing stamped assets, if we want to do that.

https://github.com/bazelbuild/bazel/issues/1025#issuecomment-194250659 implies that generated output filenames might eventually be possible, but I can't find any other issue discussing it, so I'm not holding my breath.

PR #111, opened 2017-09-30
Add rudimentary search result ranking

jboning:
Use some extremely basic heuristics to demote results that are
apparently tests or third-party code, and promote ones that seem
especially relevant.

Though this is reasonably helpful by itself, it's very much a starting
point. A couple of obvious avenues for future improvement:
 - Use ctags definition information as an input to the scoring, and
   promote matches which are definitions.
 - Raise the backend result limit to collect and score more results,
   when we can do so quickly, then return the best $n to the user.

### DIFF
--- a/src/codesearch.h
+++ b/src/codesearch.h
@@ -113,6 +113,7 @@ struct match_result {
     vector<StringPiece> context_after;
     StringPiece line;
     int matchleft, matchright;
+    int score;
 };
 
 struct file_result {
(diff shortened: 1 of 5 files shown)


nelhage:
I wonder if we should also do a similar heuristic sort on the entire corpus (or at least within repos) while indexing, since we return matches in roughly index order; That would tend to promote better matches even when we're over the search limit.

But this seems like a strong start.

Do you mind writing a test or two? Should be relatively easy to write a C++ test in `tests/` that emits a few files our-of-order in the index and asserts that they come back in the right order.

PR #112, opened 2017-10-02
Make regex, repo, and context controls sticky

### LINKED ISSUE #60: Search across several repos, or in one user-chosen repo
At Dropbox like at probably most organizations, we have a number of different repos -- a few big ones (one for most server code, one for the desktop client, etc.), and various miscellaneous ones that don't belong in those or we just haven't gotten around to merging in.

Our internal livegrep instance searches across all of our major repos and some minor ones. There's another, entirely in-house, code-search tool that searches across just one repo at a time, with a big ol' dropdown to choose which one you want.

I'm often searching to convince myself that something isn't used anymore or some bad pattern doesn't exist, or to find out *where* something is referred to from, like when some code is clearly being invoked and I can't tell how. So I really like having comprehensive search across all the repos related to a project, which is why I built our internal livegrep instance that way.

But it can also get kind of noisy, especially in navigational searches when you're trying to type just a minimal quick pattern (with the power of search-as-you-type!) to find the thing you have in mind, and you know perfectly well which repo it's in. Especially if your work is almost always in a particular repo; which for many people it is, especially so at a company that does a good job of the "monorepo" pattern that Google and Facebook are well known for and Dropbox successfully mostly moved to in 2014 and 2015, where e.g. all your web apps and backend services (and maybe more things) are seamlessly source-controlled together. So the desire to search just one repo is a pretty common reason that users give for preferring the other tool. And we're about evenly split on users who prefer that tool vs. our livegrep instance, so that's a lot of users with that preference.

Crucially, if the last time you used that other tool you searched repository X, then the next time you go to its front page and just type a quick query, you'll get a search over repository X.

I'm not sure how to integrate this well with Livegrep's UI and the all-repos search that I also value. So this is a less concrete feature request than usual. Ideas?

jboning:
Behaviour is as follows:
 - Any time you modify one of the query controls in the UI, your modification is stored in a cookie.
 - If you navigate to the blank livegrep search page, the control settings are loaded from your cookie, falling back to defaults for any controls you've never explicitly set.
 - If you navigate to a livegrep search with a query already provided, the control settings are loaded from the URL and your preferences are ignored. This way, links to search results are stable (modulo changes to the underlying index, of course).

This doesn't make the "case" control sticky because I believe "auto" really is what the user wants 99% of the time.

Fixes #60 by addressing this portion of the issue:

> Crucially, if the last time you used that other tool you searched repository X, then the next time you go to its front page and just type a quick query, you'll get a search over repository X.

### DIFF
--- a/web/src/codesearch/codesearch_ui.js
+++ b/web/src/codesearch/codesearch_ui.js
@@ -1,5 +1,6 @@
 var html = require('html');
 var Backbone = require('backbone');
+var Cookies = require('js-cookie');
 
 var Codesearch = require('codesearch/codesearch.js').Codesearch;
 var RepoSelector = require('codesearch/repo_selector.js');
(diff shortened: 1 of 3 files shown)


nelhage:
Nice

PR #114, opened 2017-10-04
Add `test` target so `bazel test //test` works

brandon-rhodes:
A bare `test_suite()` target automatically includes all tests in its `BUILD` file, allowing the tests to be run without needing to type their names.


nelhage:
Ah, nifty. Thanks, and today I learned a thing :)

PR #115, opened 2017-10-04
Stop tag file regular expressions from eating tabs

brandon-rhodes:
Since the identifier field and filename field in a tags file cannot
contain tabs, our regular expressions that match those fields should
avoid trying to eat tabs.  The actual chance of a mismatch is small, but
this tiny bit of extra correctness should also give the patterns the
best possible chance of running quickly.


nelhage:
Thanks!

PR #121, opened 2017-11-06
Add reload

brandon-rhodes:
This stand-alone command supports more complicated workflows than simply running `livegrep-fetch-reindex` and wanting an immediate reload when it finishes (for example, this lets you `livegrep-fetch-reindex`, then build a separate tags index, and only then reload).


nelhage:
:+1: belatedly

Can you fix the build, though? https://circleci.com/gh/livegrep/livegrep/32
I'll enable builds on PRs so we catch this earlier next time.

PR #126, opened 2017-11-30
Avoid double slashes in file viewer URLs

brandon-rhodes:
One of our file viewing applications would return 404 on URLs with a
double slash in them, so this helps the js code avoid building them.


tpcwang:
Seems reasonable. :+1:

PR #127, opened 2017-11-30
Bugfix: without a tags index, build no searcher

brandon-rhodes:
Several conditionals in `grpc_server.cc` expect the tags `code_searcher`
pointer to be NULL if no tag index has been provided:

```
if (tagdata != nullptr)
response->set_has_tags(tagdata_ != nullptr);
if (tagdata_ == NULL)
```

But, alas, the initialization routine always creates a tags searcher.
Even worse, it leaves it un-finalized if no tags index was provided.
Among other things, this crashes codesearch if `tags:...` is specified
but no tags index has been provided.  And with the debut of our new
tags-first search logic, it also makes the service crash for normal
searches (!) if no tags index is available.  So here we pivot to the
behavior the server code expects: a NULL pointer if no tags index is
loaded.

### DIFF
--- a/src/tools/codesearch.cc
+++ b/src/tools/codesearch.cc
@@ -109,7 +109,6 @@ void build_index(code_searcher *cs, const vector<std::string> &argv) {
 }
 
 void initialize_search(code_searcher *search,
-                       code_searcher *tags,
                        int argc, char **argv) {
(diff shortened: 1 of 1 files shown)


tpcwang:
:+1:

nelhage:
Whoops, I'm working on fixing PR CI.

PR #129, opened 2017-12-10
Fix checks of calls returning a file descriptor

brandon-rhodes:
The daemon is failing on some servers because these calls return 0,
perhaps because stdin has been closed during daemonization.
These asserts should accept all valid fd numbers, including zero.


nelhage:
This seems right. That said, even if your daemonization system closes stdio, I'd suggest configuring it to re-open `/dev/null` onto 0, 1, and 2; You'll avoid a whole host of dumb bugs like this, or other bugs where code accidentally tries to log something and ends up sending it to a random network socket, instead.

PR #132, opened 2017-12-14
Print error and exit if index cannot be opened

brandon-rhodes:
The error message that the assert was printing is somewhat opaque, and
did not record which file could not be opened:
```
codesearch: src/dump_load.cc:361: load_allocator::load_allocator(code_searcher*, const string&): Assertion `fd_ > 0' failed.
```

### DIFF
--- a/src/dump_load.cc
+++ b/src/dump_load.cc
@@ -30,7 +30,11 @@ public:
         hdr_() {
         assert(!stream_.fail());
         fd_ = open(path.c_str(), O_RDWR|O_APPEND);
-        assert(fd_ >= 0);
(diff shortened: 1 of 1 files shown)


nelhage:
We also have a [`die_errno`](https://github.com/livegrep/livegrep/blob/88e2b013f25d29fe67f697c368137b8d9531e120/src/tools/codesearch.cc#L60) you could use, although probably I'd want to upgrade it to take a `va_list` to really be useful.

nelhage:
Oh, I guess that's local to `codesearch.cc` and not in `debug.h`. Might be worth adding that globally at some point.

PR #134, opened 2017-12-15
Add a "total wallclock time" calculated from livegrep

alexmv:
This allows parsing the logs to get a close approximation of the time the end-user waited for a response.


nelhage:
:+1: 

Is this based on an old sha1 or something? Is that why circle is failing?

alexmv:
Yeah.  I've rebased, and we'll see how it likes it now.

PR #137, opened 2018-01-02
Add buttons to narrow results by file extensions

brandon-rhodes:
A popular request from users: buttons that automatically narrow the
current search to files matching a single extension.  The extensions
offered are all culled from the currently displayed files.

Because these buttons need to advance the search history URL to a new
state from which the user often wants to return by hitting Back, this
diff also fixes livegrep's previous problem of not preserving browser
URL history; we now create a browser history state for each search
that successfully returns and is drawn on the screen.

### DIFF
--- a/web/htdocs/assets/css/codesearch.css
+++ b/web/htdocs/assets/css/codesearch.css
@@ -188,8 +188,12 @@ a:hover {
     margin-top: 10px;
 }
 
-.path-results {
+.file-extensions, .path-results {
     margin-bottom: 15px;
+}
+
+.file-extensions button {
+    margin-left: 4px;
 }
 
 .file-group {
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
Happy New Year, livegrep maintainers!

I welcome feedback on whether these file-extension buttons, which were a repeatedly requested feature here at Dropbox, might also be useful to other Livegrep users and thus a candidate for inclusion in the main source tree; and also on whether they should simply always be on, as in this diff, or whether they should be a feature that administrators would want to turn on and off. Thanks for your feedback on both questions, and on any other aspects of this prototype of the feature!

nelhage:
Sorry for the belated review! I like it, and am happy to take this as a default.

PR #139, opened 2018-01-17
Free tags code_searcher before creating a new one

brandon-rhodes:
Otherwise, a server run with `-load_tags` will leak a memory map to its
tags index every time it reloads, potentially keeping old index files
alive instead of letting the filesystem remove them.


nelhage:
:+1: good catch.

We should really use a `unique_ptr` here or something; I'll follow up with something like that.

PR #141, opened 2018-01-17
Use a unique_ptr

nelhage:
cc @brandon-rhodes

### DIFF
--- a/src/tools/codesearch.cc
+++ b/src/tools/codesearch.cc
@@ -31,6 +31,7 @@
 #include <functional>
 #include <future>
 #include <thread>
+#include <memory>
 
 #include <gflags/gflags.h>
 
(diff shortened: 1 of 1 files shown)


brandon-rhodes:
Very nice! I'd almost tried a `unique_ptr` but didn't know the ramifications of calling it a "unique pointer" when, in fact, a copy was being made by the invocation of `listen_grpc()` — I would have assumed that the language would have forbidden its being copied into a parameter list? In any case, this is more elegant than my solution — thanks!

nelhage:
The `.get()` casts it out of being a `unique_ptr`, and so we're relying on `listen_grpc` not to store the pointer somewhere or otherwise reference it after return. So we don't have the same level of safety that we would in, say, Rust. 

We could do something like past a `const unique_ptr&` to the method to get greater safety, since that would make it clear that `listen_grpc` only gets a read-only view of someone else's `unique_ptr`, but I didn't feel like being that clever.

PR #142, opened 2018-01-17
If a tag index is provided, search it first

brandon-rhodes:
On the assumption that tagged lines have greater semantic weight than
other lines that match the user’s search, let’s power a search by
consulting the tags index first then falling back to the main index.

### DIFF
--- a/src/tagsearch.cc
+++ b/src/tagsearch.cc
@@ -18,21 +18,21 @@ using boost::filesystem::path;
 
 namespace {
 
-std::string create_partial_regex(RE2 *re, const char *wildstar) {
+std::string create_partial_regex(RE2 *re, const char *wildchar) {
     if (!re)
-        return wildstar;
+        return std::string(wildchar) + "+";
 
     std::string pattern = re->pattern();
 
     if (pattern.front() == '^')
-        pattern.erase(pattern.begin());
+        pattern.erase(0, 1);
     else
-        pattern.insert(0, wildstar);
+        pattern.insert(0, std::string(wildchar) + "*");
 
     if (pattern.back() == '$')
         pattern.erase(pattern.size() - 1);
     else
-        pattern.append(wildstar);
+        pattern.append(std::string(wildchar) + "*");
 
     return pattern;
 }
@@ -124,8 +124,26 @@ bool tag_searcher::transform(query *q, match_result *m) const {
 }
 
 std::string tag_searcher::create_tag_line_regex_from_query(query *q) {
-    return create_tag_line_regex(create_partial_regex(q->line_pat.get(), "[^\t]*"),
-                                 create_partial_regex(q->file_pat.get(), "[^\t]*"),
-                                 "\\d+",
-                                 create_partial_regex(q->tags_pat.get(), ".*"));
+    /* To make tag searches as efficient as possible, we return a
+       pattern that is only as long as it needs to be to specify all of
+       the query constraints.  In particular, it used to be a minor
+       disaster that a simple tags search for a 2-letter string like
+       "Ab" produced an RE that also contained the substring ";\"\t",
+       because that 3-character substring was more attractive to the
+       indexing logic than the 2-character pattern "Ab" but, alas, those
+       3 characters appeared in every single line of the tags file and
+       were therefore worthless for reducing the search space. */
+    std::string regex("^");
+    regex += create_partial_regex(q->line_pat.get(), "[^\t]");
+    regex += "\t";
+    if (q->file_pat || q->tags_pat) {
+        regex += create_partial_regex(q->file_pat.get(), "[^\t]");
+        regex += "\t";
+        if (q->tags_pat) {
+            regex += "\\d+;\"\t";
+            regex += create_partial_regex(q->tags_pat.get(), ".");
+            regex += "$";
+        }
+    }
+    return regex;
 }
--- a/src/tools/grpc_server.cc
+++ b/src/tools/grpc_server.cc
@@ -8,6 +8,7 @@
 #include "src/tools/limits.h"
 #include "src/tools/grpc_server.h"
 
+#include "gflags/gflags.h"
 #include <json-c/json.h>
 
 #include <algorithm>
@@ -23,6 +24,8 @@ using grpc::Status;
 using grpc::StatusCode;
 
 using std::string;
+
+DEFINE_int32(max_matches, 50, "The default maximum number of matches to return for a single query.");
 
 class CodeSearchImpl final : public CodeSearch::Service {
  public:
@@ -106,12 +109,12 @@ Status CodeSearchImpl::Info(ServerContext* context, const ::InfoRequest* request
     return Status::OK;
 }
 
-Status extract_regex(std::unique_ptr<RE2> *out,
+Status extract_regex(std::shared_ptr<RE2> *out,
                      const std::string &label,
                      const std::string &input,
                      bool case_sensitive) {
     if (input.empty()) {
-        out->reset(nullptr);
+        out->reset();
         return Status::OK;
     }
 
@@ -119,7 +122,7 @@ Status extract_regex(std::unique_ptr<RE2> *out,
     default_re2_options(opts);
     opts.set_case_sensitive(case_sensitive);
 
-    std::unique_ptr<RE2> re(new RE2(input, opts));
+    std::shared_ptr<RE2> re(new RE2(input, opts));
     if (!re->ok()) {
         return Status(StatusCode::INVALID_ARGUMENT, label + ": " + re->error());
     }
@@ -127,11 +130,11 @@ Status extract_regex(std::unique_ptr<RE2> *out,
     return Status::OK;
 }
 
-Status extract_regex(std::unique_ptr<RE2> *out,
+Status extract_regex(std::shared_ptr<RE2> *out,
                      const std::string &label,
                      const std::string &input) {
     if (input.empty()) {
-        out->reset(nullptr);
+        out->reset();
         return Status::OK;
     }
     bool case_sensitive = std::any_of(input.begin(), input.end(), isupper);
@@ -159,9 +162,31 @@ Status parse_query(query *q, const ::Query* request, ::CodeSearchResult* respons
 
 class add_match {
 public:
-    add_match(CodeSearchResult* response) : response_(response) {}
+    typedef std::set<std::string> line_set;
+
+    add_match(CodeSearchResult* response)
+        : unique_lines_(NULL), response_(response) {}
+
+    add_match(line_set* unique_lines, CodeSearchResult* response)
+        : unique_lines_(unique_lines), response_(response) {}
+
+    int match_count() {
+        return response_->results_size();
+    }
 
     void operator()(const match_result *m) const {
+        if (unique_lines_ != NULL) {
+            // Avoid a duplicate if a line is returned once from the
+            // tags search then again during the main corpus search.
+            std::string key = std::string(m->file->tree->name)
+              + " " + std::string(m->file->tree->version)
+              + " " + std::string(m->file->path)
+              + " " + std::to_string(m->lno);
+            bool was_inserted = unique_lines_->insert(key).second;
+            if (!was_inserted)
+                return;
+        }
+
         auto result = response_->add_results();
         result->set_tree(m->file->tree->name);
         result->set_version(m->file->tree->version);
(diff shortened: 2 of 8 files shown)


brandon-rhodes:
I realize that this is a larger-than typical change, and also that it probably reveals that this is my first time in a very long time writing any substantial amount of C++ code — so I'm prepared for it needing quite a bit of work (and perhaps also needing to make the feature optional?) before it could be considered for inclusion.

In particular, you might be able to offer a much better idea than I came up with for how to control the number of results returned; I had to move it over nearer to the code doing the de-duplication that this requires since the two indexes will have some of the same lines in common.

src/tools/grpc_server.cc
```diff
@@ -235,8 +290,58 @@ Status CodeSearchImpl::Search(ServerContext* context, const ::Query* request, ::
...
+
+            q.max_matches = max_matches - cb.match_count();
+            if (q.max_matches > 0) {
+
+              /* Third and final pass: full corpus search. */
+              q.line_pat.reset(new RE2(line_pat, q.line_pat->options()));
+              code_searcher::search_thread *search;
+              if (!pool_.try_pop(&search))
+                search = new code_searcher::search_thread(cs_);
+              search->match(q, cb, cb, &stats);
```
nelhage: Hm, this is maybe minor, but we're going to lose accuracy on `stats` here; the last search will win. I think the thing I might suggest is:

- add a default constructor for `stats` that zeros out all the fields
- delete the [`memset`](https://github.com/livegrep/livegrep/blob/0eeba4bcbd7d73f75288dc46bd9cfebb1c887196/src/codesearch.cc#L1150) here
- make [this code](https://github.com/livegrep/livegrep/blob/0eeba4bcbd7d73f75288dc46bd9cfebb1c887196/src/codesearch.cc#L1166-L1174) and [this code](https://github.com/livegrep/livegrep/blob/0eeba4bcbd7d73f75288dc46bd9cfebb1c887196/src/codesearch.cc#L246-L252) *add* to `stats`, instead of setting.

that will let us accumulate counters with no additional work

src/tools/grpc_server.cc
```diff
@@ -235,8 +290,58 @@ Status CodeSearchImpl::Search(ServerContext* context, const ::Query* request, ::
...
+        add_match cb(&unique_lines, response);
+
+        /* To surface the most important matches first, start with tags.
+           First pass: is the pattern an exact match for any tags? */
+        regex = "^" + line_pat + "$";
+        q.line_pat.reset(new RE2(regex, q.line_pat->options()));
+        run_tags_search(q, tagdata_, cb, tagmatch_, stats);
+
+        q.max_matches = max_matches - cb.match_count();
+        if (q.max_matches > 0) {
```
nelhage: Let's extract this whole inner bit into a function, and then instead of these continually-nesting `if (q.max_matches > 0)` clauses, we can  `if (q.max_matches <= 0) return` and keep the indentation under control.

src/tools/grpc_server.cc
```diff
@@ -159,9 +162,31 @@ Status parse_query(query *q, const ::Query* request, ::CodeSearchResult* respons
...
+
+    int match_count() {
+        return response_->results_size();
+    }
 
     void operator()(const match_result *m) const {
+        if (unique_lines_ != NULL) {
+            // Avoid a duplicate if a line is returned once from the
+            // tags search then again during the main corpus search.
+            std::string key = std::string(m->file->tree->name)
```
nelhage: It's prooobably fine here but doing a lot of string concatenation makes me kind of sad; it's got poor performance (probably fine in practice since these things are mostly short), and it could go wrong if `name` or `version` have spaces.

Let's instead store a `std::pair<indexed_file*,int>`, which should uniquely identify lines (there's only one `indexed_file` per file, so comparing by address is fine). That will be both shorter, faster, and more accurate.

nelhage: Especially if we make that change, this should be ~free both in terms of performance and in terms of having no impact in the normal case, so I'd make this deduplication unconditional, store just a `std::set` instead of a pointer to one, and simplify the code thereby.

src/tagsearch.cc
```diff
@@ -124,8 +124,26 @@ bool tag_searcher::transform(query *q, match_result *m) const {
 }
 
 std::string tag_searcher::create_tag_line_regex_from_query(query *q) {
-    return create_tag_line_regex(create_partial_regex(q->line_pat.get(), "[^\t]*"),
-                                 create_partial_regex(q->file_pat.get(), "[^\t]*"),
-                                 "\\d+",
-                                 create_partial_regex(q->tags_pat.get(), ".*"));
+    /* To make tag searches as efficient as possible, we return a
```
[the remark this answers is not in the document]
brandon-rhodes: I will turn my attention to the C++ tests and see if I can provide some coverage for this function!

src/tools/grpc_server.cc
brandon-rhodes: :+1:

src/tools/grpc_server.cc
brandon-rhodes: Oh, nice, I was indeed bothered by the creeping indentation and that will solve it. It wasn't clear to my eyes that the argument list wouldn't need to be horrendously long to provide all the context used here, but I'll see what it looks like.

src/tools/grpc_server.cc
brandon-rhodes: :+1:

src/tools/grpc_server.cc
brandon-rhodes: @nelhage Attempting to store the `std::set` on the class itself runs afoul of the `const` declaration of `void operator()(const match_result *m) const {`. The `const` seems to be there because the callback will later be coerced to a `std::function<void (const struct match_result*)>` which, I am told by @jboning, makes a copy of the whole struct, and it would be the copy’s `set` that received the lines instead of the original struct's `set`.

I'll go ahead and attempt to improve the type of the set so we can avoid the string concatenation, though!

nelhage: Ah yeah, that makes sense, I forgot that we were passing these by value. I think storing a reference or pointer to external storage makes sense then. But yeah, let's still improve the type.

brandon-rhodes:
@nelhage All right! I have attempted a first pass at each of your ideas — each of which indeed seems to have improved the code and simplified things! Let me know how it looks; I'm happy to do a few more round trips if there are places where you can show me how to make the C++ code more idiomatic, as it would be good practice for me.

src/tools/grpc_server.cc
```diff
@@ -165,26 +165,22 @@ class add_match {
     typedef std::set<std::string> line_set;
 
     add_match(CodeSearchResult* response)
-        : unique_lines_(NULL), response_(response) {}
-
-    add_match(line_set* unique_lines, CodeSearchResult* response)
-        : unique_lines_(unique_lines), response_(response) {}
+        : unique_lines_(new line_set), response_(response) {}
```
nelhage: This is a memory leak, since no one frees it. I think stack-declaring it in the parent and passing it in reference is probably the best option.

src/tools/grpc_server.cc
```diff
@@ -174,14 +175,10 @@ class add_match {
     void operator()(const match_result *m) const {
         // Avoid a duplicate if a line is returned once from the
         // tags search then again during the main corpus search.
-        std::string key = std::string(m->file->tree->name)
-            + " " + std::string(m->file->tree->version)
-            + " " + std::string(m->file->path)
-            + " " + std::to_string(m->lno);
-        bool was_inserted = unique_lines_->insert(key).second;
-        if (!was_inserted) {
+        line_set_key k(m->file, m->lno);
```
nelhage: You can just inline this into `unique_lines_->insert(make_pair(m->file, m->lno))` and skip the declaration.

src/tools/grpc_server.cc
```diff
@@ -177,8 +178,10 @@ class add_match {
         // tags search then again during the main corpus search.
         line_set_key k(m->file, m->lno);
         bool already_inserted = ! unique_lines_->insert(k).second;
-        if (already_inserted)
+        if (already_inserted) {
+            log("DUP");
```
nelhage: clean up this log statement

src/tools/grpc_server.cc
```diff
@@ -243,6 +246,40 @@ static std::string pat(const std::shared_ptr<RE2> &p) {
     return p->pattern();
 }
 
+void CodeSearchImpl::TagsFirstSearch_(::CodeSearchResult* response, query& q, string& line_pat, match_stats& stats) {
```
nelhage: Let's make this a `const string &` to make it clear we're not mutating. I might even just pass in `q` and extract it again via `q.line_pat->pattern()` to drop a parameter.

src/tools/grpc_server.cc
```diff
@@ -243,6 +246,40 @@ static std::string pat(const std::shared_ptr<RE2> &p) {
...
+
+    add_match cb(response);
+
+    /* To surface the most important matches first, start with tags.
+       First pass: is the pattern an exact match for any tags? */
+    regex = "^" + line_pat + "$";
+    q.line_pat.reset(new RE2(regex, q.line_pat->options()));
+    run_tags_search(q, tagdata_, cb, tagmatch_, stats);
+
+    q.max_matches = max_matches - cb.match_count();
```
nelhage: `q.max_matches -= cb.match_count()` would be more idiomatic and I think clearer

nelhage:
Nice! A few more small nitpicks, and then I'm happy!

src/tools/grpc_server.cc
brandon-rhodes: What if `unique_lines_` were declared a `unique_ptr` instead of a bare old `*` — would that guarantee a free when the `add_match` object is removed from the stack?

src/tools/grpc_server.cc
brandon-rhodes: Neat! How does the destructor wind up getting called on the pair in that case, though? I'd created the declaration so that I could build it on the stack and thus know when it got deallocated.

src/tools/grpc_server.cc
brandon-rhodes: Thanks, done!

src/tools/grpc_server.cc
brandon-rhodes: I'll make it `const` right now!

As for calling `pattern()` again — I didn't see any indication of whether `pattern()` is expensive or not (I didn't check whether it has to do something like rebuild the RE's string from an introspection of its bytecode or something), hence my temptation to go ahead and pass the value while I had it already. Is it actually a cheap operation?

src/tools/grpc_server.cc
brandon-rhodes: I'll improve the naming here! This isn't a decrement; it's a throwing out of the old value of `q.max_matches` because our match count is new. I'll switch `max_matches` to `original_max_matches` to hopefully make things clear.

src/tools/grpc_server.cc
jboning: I think trying to use a `unique_ptr` in `add_match` is what led me to discover the terrible truth about `std::function` a long time ago: when an `add_match` is used as a `std::function`, a new `std::function` object is constructed, with a copy of the `add_match` object. But `unique_ptr` has no copy constructor, so using a `unique_ptr` as a field means that `add_match` also can't be copy-constructed, and so it fails at compilation time.

brandon-rhodes:
Thanks for the additional ideas! Tests are passing here locally on my latest update, so let's see what the CI says.

src/tools/grpc_server.cc
nelhage: https://github.com/google/re2/blob/master/re2/re2.h#L264

it's stored directly internally, which is what I'd expect -- the bytecode is going to be a lossy translation, since it canonicalizes, and I'd expect `pattern()` to return the input unmodified.

src/tools/grpc_server.cc
nelhage: Temporaries are constructed on the stack and destructed at the end of the containing statement, so just after the call to `insert`. It's a fairly common C++ idiom to use that fact in situations like this.

src/tools/grpc_server.cc
nelhage: Yep, @jboning is right. A `shared_ptr` would potentially work with approximately the desired semantics.

src/tools/grpc_server.cc
brandon-rhodes: Thanks for tracking down the line number! (I hadn't had time to yet today.) I'll go remove that parameter —

nelhage:
It looks like we're still leaking the `unique_lines_` set. If we can clean that up -- tbh, let's just make it a `shared_ptr`; that feels slightly sloppy but will be easy and safe -- I'm happy to take!

brandon-rhodes:
You were right: the `line_set` was best extracted as a completely separate automatic. My brief attempt to make it a `unique_ptr` exploded because of the same problem that I ran into earlier: the way we pass `add_match` as a callback does a copy, and you're not allowed to copy a `unique_ptr`.

brandon-rhodes:
OH WAIT — I just re-read your most recent message, now that I'm more awake. You said "shared" pointer, not unique.

But my guess is that it'll be cleaner to just stick with the second automatic variable?

nelhage:
Thanks! Yeah, I was kind of torn here between dueling heuristics of "If a variable logically has automatic scope, prefer to avoid a heap allocation" and "prefer managed pointers to carrying around bare pointers", so I'm fine either way.

PR #144, opened 2018-02-05
Avoid spamming browser history with each keystroke

brandon-rhodes:
Users have complained that, following my `1705f305c15c` contribution,
Livegrep is now a bit aggressive about spamming their browser history —
in cases where searches return quickly, every single character they type
creates a new history item that they have to slog "Back" through if they
want to return to the previous page.  So let’s back off and not create
new history items while they’re clearly in the middle of typing.

### DIFF
--- a/web/src/codesearch/codesearch_ui.js
+++ b/web/src/codesearch/codesearch_ui.js
@@ -9,6 +9,7 @@ function init(initData) {
 "use strict";
 
 var h = new html.HTMLFactory();
+var last_url_update = 0;
 
 function vercmp(a, b) {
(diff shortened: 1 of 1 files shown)


nelhage:
This is a separate debounce because even with our debounce and only with updating on result, we're still too fast for people? That's not surprising, I guess -- the existing debounce regularly updates several times even as I type. Cool.

brandon-rhodes:
> we're still too fast for people?

I know, I was surprised too!

But the developer who showed me the screenshot, who must be much closer to the search server than I am on the network, had a history entry for every single keystroke. Every one!

So either they're on a very fast network or a very slow typist. :)

PR #145, opened 2018-02-12
Add [y] shortcut to create permalink

brandon-rhodes:
Many of our users asked us to better expose the support that the
Livegrep file browser offers for permalinks to exact revisions of a
particular file, so let’s add links between HEAD and the permalink.


nelhage:
Can you fix the tests and I'll take a look?

brandon-rhodes:
Fixed! I'll go finally read now about how to get Emacs to run gofmt when I hit Save.

PR #148, opened 2018-02-14
Add support for walking a custom-ordered subset of filesystem paths

stappon:
This lets codesearch produce an index that has been populated in a specific order to position more interesting files first, or to exclude uninteresting files entirely.

I decided to go with this approach rather than implementing ranking inside of livegrep because it gives the user the flexibility to write arbitrarily complicated filtering/ranking logic while keeping livegrep generic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ config.json
 vendor/re2/obj/libre2.a
 /bazel-*
 compile_commands.json
+.devbox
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@ config.json
 vendor/re2/obj/libre2.a
 /bazel-*
 compile_commands.json
+.devbox
```
nelhage: I'm happy to take this, but fyi you can also use `.git/info/exclude` to get un-versioned local excludes, or set `core.excludesfile` in `~/.gitconfig` to populate a user-local exclude.

PR #156, opened 2018-03-19
Let '/' search for selected text on main page

brandon-rhodes:
Another popular request from among our most dedicated Livegrep users!
This duplicates a bit of code from `fileview.js` but as it looked like
the two have been kept separate so far, I didn't choose this diff as the
moment to create a common module where the shared code could live.

### DIFF
--- a/web/src/codesearch/codesearch_ui.js
+++ b/web/src/codesearch/codesearch_ui.js
@@ -592,6 +592,12 @@ var MatchesView = Backbone.View.extend({
       event.preventDefault();
       if (CodesearchUI.input_regex.is(':checked'))
         t = t.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // from mozilla docs
+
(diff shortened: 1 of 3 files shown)


web/src/codesearch/codesearch_ui.js
```diff
@@ -570,6 +579,28 @@ var MatchesView = Backbone.View.extend({
       q = 'file:' + ext + ' ' + q;
     CodesearchUI.input.val(q);
     CodesearchUI.newsearch();
+  },
+  _handleKey: function(e) {
+    if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey)
+      return;
+    var which = event.which;
+    console.log(KeyCodes);
```
nelhage: Remove the debug print, please

brandon-rhodes: Argh!!!

PR #161, opened 2018-03-21
Fixes following a Bazel upgrade on our end

### LINKED ISSUE #160: Attempting build results in "no such remote ref"
For some reason Bazel's attempt to fetch the pinned revision of `com_github_nelhage_boost` is resulted in an error, even though I think that ref does exist on the remote:
```
$ bazel build //...
ERROR: error loading package '': Encountered error while reading extension file 'boost/boost.bzl': no such package '@com_github_nelhage_boost//': Traceback (most recent call last):
        File "/home/brhodes/.cache/bazel/_bazel_brhodes/7f6357a54daf9c6f2234f2479b3b3fc6/external/bazel_tools/tools/build_defs/repo/git.bzl", line 69
                _clone_or_update(ctx)
        File "/home/brhodes/.cache/bazel/_bazel_brhodes/7f6357a54daf9c6f2234f2479b3b3fc6/external/bazel_tools/tools/build_defs/repo/git.bzl", line 44, in _clone_or_update
                fail(("error cloning %s:\n%s" % (ctx....)))
error cloning com_github_nelhage_boost:
+ cd /home/brhodes/.cache/bazel/_bazel_brhodes/7f6357a54daf9c6f2234f2479b3b3fc6/external
+ rm -rf /home/brhodes/.cache/bazel/_bazel_brhodes/7f6357a54daf9c6f2234f2479b3b3fc6/external/com_github_nelhage_boost
+ git clone --depth=1 https://github.com/nelhage/rules_boost /home/brhodes/.cache/bazel/_bazel_brhodes/7f6357a54daf9c6f2234f2479b3b3fc6/external/com_github_nelhage_boost
Cloning into '/home/brhodes/.cache/bazel/_bazel_brhodes/7f6357a54daf9c6f2234f2479b3b3fc6/external/com_github_nelhage_boost'...
+ cd /home/brhodes/.cache/bazel/_bazel_brhodes/7f6357a54daf9c6f2234f2479b3b3fc6/external/com_github_nelhage_boost
+ git reset --hard d6446dc9de6e43b039af07482a9361bdc6da5237
fatal: Could not parse object 'd6446dc9de6e43b039af07482a9361bdc6da5237'.
+ git fetch --depth=1 origin d6446dc9de6e43b039af07482a9361bdc6da5237:d6446dc9de6e43b039af07482a9361bdc6da5237
error: no such remote ref d6446dc9de6e43b039af07482a9361bdc6da5237
```

brandon-rhodes:
We seem to be on a version of Bazel that now performs depth 1 checkouts
but that does not yet incorporate:

https://github.com/bazelbuild/bazel/commit/6496b1d1d25025f33406b1eaf9949cab126f08bd

that would presumably let us make our checkouts a bit less shallow, and
thus pin to arbitrary commits.  Instead, we must at the moment pin to
actual tags and branches on GitHub.  This diff makes the changes
necessary to pin to available versions instead of arbitrary commits;
tweaks a few things that broke during the necessary upgrades this
entailed; and where appropriate, follows the current Bazel docs in
replacing git repository fetches with HTTP archives.

The one thing I don’t like is specifying “master” for the repository
“org_pubref_rules_protobuf” — should we instead ask them to make a more
recent release?  Or fork the repository ourselves as a target?

This fixes #160 upstream.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -5,7 +5,6 @@ go_prefix("github.com/livegrep/livegrep")
 
 compilation_database(
     name = "compilation_db",
-    exec_root_marker = True,
     targets = [
         "//src/tools:codesearch",
         "//src/tools:codesearchtool",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -8,7 +8,7 @@ load(
 
 git_repository(
     name = "org_pubref_rules_protobuf",
-    commit = "ff3b7e7963daa7cb3b42f8936bc11eda4b960926",
+    commit = "master",
     remote = "https://github.com/pubref/rules_protobuf",
 )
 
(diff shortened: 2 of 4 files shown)


brandon-rhodes:
I realize that this might not be merge-able because (a) it might not work with your Bazel, and (b) you might just prefer to wait until `shallow_since` becomes available in all Bazels so that none of the hashes that Livegrep has pinned have to move.

But I figured I should get this up here in any case, at least for users who might be stuck with a similar Bazel version to ours!

brandon-rhodes:
I think this is now ready for review! The upstream project `rules_protobuf` added a tag so there are now no references to branches like `master` in the workspace file.

nelhage:
I'm still wary to land this since I prefer sha1's, since those are guaranteed to be stable over time. 

It looks like the bazel issue is fixed in 0.12; Is this still a problem for you? If so, and if it's helpful, I could be talked into landing this temporarily.

brandon-rhodes:
I agree that hashes are semantically far superior to names, so I'll close this pull request for now; hopefully in a few months our Bazel will be upgraded. In the meantime, I'll re-open if we have anything else to add!

PR #164, opened 2018-05-21
Switch to Prism.js syntax highlighting

brandon-rhodes:
Our users have reported several problems with “highlight.js” syntax
highlighting, and the project does not appear to be addressing its open
issues on GitHub quickly. Switching to “prism.js” seems to have
addressed the biggest pain points.

Also, went ahead and added a few more file types for syntax highlighting.


brandon-rhodes:
You might, of course, recommend another direction for fixing these problems — but this is what I've done locally for the moment to try getting things fixed.

nelhage:
I'm not super familiar with the space of syntax highlighters, so if this seems to work better, let's goferit

gnprice:
For my own awareness when dropping in syntax highlighters elsewhere, I am curious what kind of issues your users were hitting! Was the highlighting inaccurate, did it run slowly, or something else?

brandon-rhodes:
The highlighting was inaccurate — for some languages (if I get time I'll look up which) highlighting would work through part of the file, then the whole rest of the file would be a different color.

alexmv:
Handled single-quotes in comments wrong, mostly.

gnprice:
Good to know, thanks!

The highlight.js authors/maintainers talk a stern game about how a code highlighter needs to be accurate most of all, and they're willing to sacrifice some speed and code size for that. Which I like in principle, but it seems not to have totally worked out that way.

There's perhaps an interesting architectural lesson here: prism.js is regexes, which inherently have a lower ceiling on accuracy than highlight.js's more general parsers do -- but they're a lot more accessible to hack on. Perhaps that contributes to issues in highlight.js languishing unfixed, while prism.js I guess achieves something closer to its design's inherent ceiling.

PR #165, opened 2018-05-23
Add max_matches as a query hint

tpcwang:
This helps with discoverability and ease of use.


nelhage:
Sure! We may want end up wanting some UI if you're over the server-configured maximum, but as long as we have this might as well offer it.

PR #168, opened 2018-06-08
Add “ScriptNonce” that adds attribute to scripts

brandon-rhodes:
Here at Dropbox we’ll be using a nonce to protect each bare `<script>`
tag, and this code makes it easy for third parties to add nonces without
editing the code in too many places.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -96,7 +96,7 @@ func (s *server) ServeSearch(ctx context.Context, w http.ResponseWriter, r *http
 	}
 	s.renderPage(w, &page{
 		Title:         "code search",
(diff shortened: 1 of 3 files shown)


nelhage:
Looks like more `gofmt` errors :) lgtm otherwise, feel free to `gofmt` and merge yourself.

brandon-rhodes:
I have no idea how Emacs and gofmt conspired to add those spaces when I hit "Save". I've just hit "Save" again so we'll see if CI is happier this time!

brandon-rhodes:
Good, it's happier :)

PR #171, opened 2018-06-14
Make template relationships explicit in their text

brandon-rhodes:
Instead of composing templates by first rendering the page template to a
string and then inserting it into the layout template, let’s load real
Go template stacks so that the relationship is explicit inside each page
template, and so that only a single rendering run is necessary.

By letting all templates render in the same data context, this will open
the way to CSP noncing of `<script>` tags whether they’re in the layout
or whether they live down in a page template.


brandon-rhodes:
I will modestly tout it as an advantage of this pull request that it not only eliminates a `.go` file, but that — to make it easier to have the project open in an editor — it drops the number of Go files named `templates.go` from 2 to 1. :)

nelhage:
Haha. Those are both great reasons to like this PR. 

Sorry for the delayed review, but this looks great!

PR #172, opened 2018-06-26
Fix nonces by adding them to all script, link tags

brandon-rhodes:
It turns out?  Our CSP wants nonces on *every* script and link tag, not
just inline scripts.  Thanks to the recent template simplification, it’s
now easy to use the Nonce both up in the main “layout.html” template and
also down in specific templates like “fileview.html”.

This also simplifies the code by precomputing the full text of the HTML
attribute, so that the templates needn’t be festooned with “if” statements.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -22,13 +22,13 @@ import (
 
 type page struct {
 	Title         string
-	ScriptNonce   string
 	ScriptName    string
 	ScriptData    interface{}
 	IncludeHeader bool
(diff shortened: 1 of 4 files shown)


brandon-rhodes:
(Oh, and: the new argument `r` to the page rendering logic is necessary because, in our case, we compute the nonce using some information in the request.)

brandon-rhodes:
Thanks! :)

PR #173, opened 2018-07-10
Switch sparseconfig.h to an explicit dep

fowles:
Instead of using a patch file and special rule.  Use an ordinary build
rule and inject the deps with the correct path.


fowles:
Hey Nelson, I ended up using some tricks from your other BUILD files to get a small project working.  I also found a slightly simpler regime for making sparsehash work (no need for a custom build rule).  So I figure I would offer it up to you.  Feel free to ignore reject if you don't want to bother.

fowles:
Updated to actually work ;)

nelhage:
Nice, thanks! This is a nice cleanup.

PR #174, opened 2018-07-19
Support for jump to definition using language servers

ilinum:
This PR adds support for jump to definitions when using built-in fileview. 

This is done using [language servers](https://microsoft.github.io/language-server-protocol/), making all the code in this PR language agnostic (we tried it with [Python](https://github.com/palantir/python-language-server) and [Go](https://github.com/sourcegraph/go-langserver)).

In order to use the feature, you have to start a language server on a certain port and then specify it in the config.

For example, if you used a go language server for livegrep, you would first start it on any port (for this example, let's say it's `1234`), and have the following `index.json`:
```
{
    "name": "livegrep",
    "repositories": [
        {
            "name": "livegrep",
            "path": "/livegrep",
            "revisions": [ "HEAD" ],
            "metadata": {
                "github": "nelhage/livegrep",
                "url-pattern": "https://github.com/{name}/blob/HEAD/src/{path}#L{lno}",
                "blame": "git"
            },
            "lang_servers": [
                {
                    "address": "0.0.0.0:1234",
                    "extensions": [
                        ".go"
                    ]
                }
            ]
        }
    ]
}
```

This project was worked on by @anuragbakshi, @hugabor, @ilinum, @ldanilek, @permutohedra, @Utsav2, @xiaovictoria8


server/langserver/langserver.go
```diff
@@ -0,0 +1,76 @@
...
+	client = &langServerClientImpl{
+		rpcClient: rpcConn,
+	}
+	return
+}
+
+func (ls *langServerClientImpl) Initialize(ctx context.Context, params *InitializeParams) (result InitializeResult, err error) {
+	err = ls.invoke(ctx, "initialize", params, &result)
+	if err != nil {
+		ls.invoke(ctx, "initialized", nil, nil)
```
ains-stripe: It seems that "initialized" is only invoked if there is an error on the call to initialize, also the error from invoking "initialized" isn't propagated back here if one were to occur

server/server.go
```diff
@@ -268,6 +291,107 @@ func (h *reloadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		writeError(ctx, w, 400, "bad_arguments", err.Error())
+		return
+	}
+
+	l := langserver.ForFile(repo, docPositionParams.TextDocument.URI)
+	langServer := s.langsrv[l.Address]
+	if langServer == nil {
+		// no language server
+		writeError(ctx, w, 404, "not_found", err.Error())
```
ains-stripe: err would be nil here, I think we want an error message string

server/langserver/langserver.go
ains-stripe: Actually looked into this a little more, and I think you want to use 
`Notify` instead of `Call` when invoking `initialized` as it's a notification not an RPC call
https://sourcegraph.com/github.com/sourcegraph/jsonrpc2/-/blob/jsonrpc2.go#L455

web/src/fileview/fileview.js
```diff
@@ -211,6 +211,138 @@ function init(initData) {
...
+    hoveringNode = null;
+  }
+
+  function hoverOverNode(node) {
+    node.className = 'hovering';
+    hoveringNode = node;
+  }
+
+  function checkIfHoverable(node) {
+    var info = getFileInfo();
```
ains-stripe: Can you use initData here for getting the repoName instead of duplicating the URL parsing logic in JS?

(the assumption that the repo name doesn't have slashes in it is also broken for GitHub repositories)

Utsav2:
@ains-stripe could you ptal? The test failures don't look related, afaict.

server/langserver/langserver.go
```diff
@@ -0,0 +1,91 @@
...
+func (ls *langServerClientImpl) Initialize(
+	ctx context.Context,
+	params *InitializeParams,
+) (InitializeResult, error) {
+
+	var result InitializeResult
+	err := ls.call(ctx, "initialize", params, &result)
+	if err != nil {
+		// it could presumably be because we're already initialized.
+		nerr := ls.notify(ctx, "initialized", nil)
```
ains-stripe: If i'm not mistaken you want to call `initialized` when `initialize` succeeds, this notifes when `initialize` fails

ilinum:
@ains-stripe ping? :)

ains-stripe:
Hey!

Sorry, currently on vacation I’ll set a reminder to take a look when I’m
back.

Just as an FYI, I have a modified version of this patch running at Stripe
that also added some connection pooling and retries as our language server
wasn’t always the most stable.

It might be worth combining efforts when I’m back, as I think it’s better
to assume the language server connection won’t last forever.

ilinum:
Oh, this is great! Yep, pooling resources sounds like a good idea.

ilinum:
ping? @ains-stripe

ains-stripe:
Hey @ilinum 

Sorry for the delay, we're currently not running this at stripe because we're still doing work on our language server but whilst I was testing it I did work on a few patches to fix some bugs that we experienced that are probably worth applying here before merging this in.

My changes are on this branch:
https://github.com/livegrep/livegrep/commits/ains-lsp-viewable

I added connection pooling here 
https://github.com/livegrep/livegrep/commit/888e502168f15e3373b39c81dc8cc7317a7db81d

General idea being that we'll grab a connection from the pool per request, recreating connections if they break, because the language server is unavailable.

There are also a few fixes for behaviour when jumping around within a file, and fixing glitches when you double-click a definition.

I haven't tried LSP with this current branch recently so I'm not sure if you've fixed them already, but i'll try get it running in our staging environment again.

Utsav2:
@ains-stripe those are great improvements! AFAIK, we haven't made any changes in our branch to fix those.

ilinum:
@ains-stripe like Utsav said, we haven't made any changes. The changes you made sounds like the way to go! 
However, I would like to actually deploy it at Dropbox and see what breaks and then fix those things. We were able to push from a branch a few months ago when we opened the PR but it involved some hacks. I would like to deploy this properly. What do we need to get this merged?

ains-stripe:
if think if you resolve the merge conflicts I can review this PR as-is and we can work from there towards something which is a little more robust

PR #177, opened 2018-07-27
 Fix internal file viewer for repository names containing '/' character

ains-stripe:
At the moment we can't use the internal file viewer because it has the assumption that repository names cannot contain forward slashes.

This isn't true of GitHub repositories ("livegrep/livegrep"), so to make the internal viewer work for repositories containing slashes it seems like the easiest thing to do is replace the route parsing logic with our own.

We do this by building a regular expression which can match on any of the configured repositories followed by a `/` and the path. I haven't thought too much about the performance implications of the regular expression with a sufficiently large number of repositories, so open to doing the matching differently.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -23,6 +23,8 @@ import (
 	"github.com/livegrep/livegrep/server/templates"
 )
 
+var serveUrlParseError = fmt.Errorf("failed to parse repo and path from URL")
+
 type page struct {
 	Title         string
 	ScriptName    string
@@ -118,13 +120,11 @@ func (s *server) ServeSearch(ctx context.Context, w http.ResponseWriter, r *http
 }
 
 func (s *server) ServeFile(ctx context.Context, w http.ResponseWriter, r *http.Request) {
-	matches := s.serveFilePathRegex.FindStringSubmatch(pat.Tail("/view/", r.URL.Path))
-	if len(matches) == 0 {
-		http.Error(w, "failed to parse repo and path given URL", 500)
+	repoName, path, err := getRepoPathFromURL(s.serveFilePathRegex, r.URL.Path)
+	if err != nil {
+		http.Error(w, err.Error(), 400)
 		return
 	}
-
-	repoName, path := matches[1], matches[2]
 
 	commit := r.URL.Query().Get("commit")
 	if commit == "" {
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -312,16 +323,39 @@ func New(cfg *config.Config) (http.Handler, error) {
 		srv.bkOrder = append(srv.bkOrder, be.Id)
 	}
 
+	var repoNames []string
 	for _, r := range srv.config.IndexConfig.Repositories {
 		srv.repos[r.Name] = r
+		repoNames = append(repoNames, r.Name)
 	}
 
+	// Sort in descending order of length so most specific match is selected by regex engine
```
nelhage: Let's refactor this logic into `func buildRepoRegex(names []string) regexp.Regexp` and write some tests.

nelhage:
Seems like a credible approach. I'd love a test for that string-building code since it seems easy and slightly hairy. I'm not overly worried about perf here, I think.

ains-stripe:
donezo @nelhage

server/server_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+func TestRepoRegexParsing(t *testing.T) {
+	repoNames := []string{"test-repo", "test-org/test-repo", "test-repo-2", "foobar"}
+
+	repoRegex, err := buildRepoRegex(repoNames)
+	if err != nil {
+		t.Errorf("unexpected error building repo regex (%v)", err.Error())
+	}
+
+	assertRepoPath(t, repoRegex, "/view/test-repo/path/to/foobar.css", "test-repo", "path/to/foobar.css", nil)
```
nelhage: This is fine as-is, but in Go I'd consider it more idiomatic to write this

```go
cases := []struct{
  url string
  repo, path string
 err error
}{
  {"/view/test-org/test-repo/path/to/foobar.css", "test-org/test-repo", "path/to/foobar.css", nil},
  {"/view/test-repo-2/path/to/foobar.css", "test-repo-2", "path/to/foobar.css", nil},
// …
}
for _,tc := range cases {
 // inline assertRepoPath here using tc.url, tc.repo, tc.path, tc.error
}
```

nelhage:
Thanks!

PR #183, opened 2018-09-02
Bump a bunch of dependencies in bazel

### LINKED ISSUE #182: Version of grpc being used?
How can I tell what version of grpc is being used when I build livegrep? I'm not too familiar with how the golang repositories work. The output I see when building contains this:

```
Downloading https://github.com/grpc/grpc/archive/ca87867579580928ca4c9fdf97051fa25bf1d386.zip via codeload.github.com: 4,818,944 bytes
```

but that's a really old version and doesn't correspond to any tagged release in the grpc repository. I also see other output that indicates it's downloading grpc from the golang repos but there's no versioning there.

The larger context around my question is that I'm querying the livegrep server from some python code which using `grpcio==1.14.1` and getting some deserialization errors that I suspect are version mismatches. But I don't know which version to use on the python side because I can't tell what version is being used to serialize on the livegrep side. Help would be appreciated, thanks!

nelhage:
fixes #182


nelhage:
cc @ains-stripe @jboning @brandon-rhodes this makes some changes to the packaging and requires a new bazel, just as a heads-up.

PR #184, opened 2018-09-02
Update protobuf

staktrace:
This updates protobuf to the latest version, although I can't tell what version of grpc is being used, and it doesn't seem to fix the deserialization problem I was having in #182.


staktrace:
Never mind, I see #183 now

PR #188, opened 2018-09-18
Add timestamp to C++ log messages

brandon-rhodes:
It has made the service difficult to manage that its log messages lack a time.


brandon-rhodes:
I have moved on from Dropbox to new projects, so won't be pursuing this improvement. I'll close it to get it off of your dashboard. Thanks for all the reviews over the years!

PR #189, opened 2018-10-03
Stop filename match from returning same name twice

brandon-rhodes:
Users reported that as the filename list grew small, they often would be
given the same filename twice. After reproducing the situation by hand,
I discovered the filename matcher needed to be careful about not
returning the same name twice in a row.

### DIFF
--- a/src/codesearch.cc
+++ b/src/codesearch.cc
@@ -426,6 +426,7 @@ void filename_searcher::operator()()
 
     // moving the left bound as we go isn't a big-O improvement, but may help a little bit.
     auto left_bound = cc_->filename_positions_.begin();
+    int previous_first = -1;
 
     for (int i = 0; i < count; i++) {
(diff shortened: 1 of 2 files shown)


nelhage:
How easy is this to reproduce? Can we easily write a test in our googletest suite?

If so, a test would be great; Also looks like it's failing tests somehow. Otherwise this looks good.

nelhage:
oh, the tests are something breaking in my gofmt/bazel hack :/ I'm traveling but I might be able to look into that this weekend or so.

brandon-rhodes:
A test might be possible, I'll have to look at the existing tests to remember what they look like.

It's extremely easy to reproduce; about half of the searches I run seem to dup a filename. I just got one a moment ago, for example.

nelhage:
Awesome, thanks.

PR #191, opened 2018-10-05
Correctly highlight lines from tags matches

brandon-rhodes:
Before we were highlighting the entire symbol returned from a tags
search, but now we’ll just highlight the actual search term for the
frequent cases where the user has typed only a prefix of the symbol.

### DIFF
--- a/src/tagsearch.cc
+++ b/src/tagsearch.cc
@@ -13,6 +13,8 @@
 #include <utility>
 #include <boost/filesystem.hpp>
 
+#include "utf8.h"
+
 using re2::RE2;
 using boost::filesystem::path;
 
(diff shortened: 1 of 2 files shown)


brandon-rhodes:
Before:

image

After:

image

brandon-rhodes:
This also simplifies how the tags search works. Instead of repeatedly mutating the original query and then having to restore its state at the end, it now passes its specialty RE through a second argument.

nelhage:
👍  with a rebase on #192

brandon-rhodes:
It's now atop 192 and it looks like tests are passing! :)

nelhage:
btw, I believe I've given you merge permission, and please feel free to click the button yourself in a case like this! Thanks for all the contributions back.

nelhage:
Ah, my bad, I hadn't! I've sent you an invite. I'll continue to review PRs, but I'm fine with merges if I've asked for trivial feedback, or if a coworker feels comfortable doing a review!

brandon-rhodes:
Wow, thanks! :)

PR #192, opened 2018-10-05
Fix gofmt in CI

nelhage:


nelhage:
cc @brandon-rhodes if you rebase your PRs on top of master after this merge I think they'll go green

brandon-rhodes:
Thanks, I'll go do that! Could this become part of `bazel test //...` so that I can learn about it locally rather than through CI?

nelhage:
Yeah, with a bit of work. I've seen a few projects implement it in various ways. e.g. we might be able to import some version of the hacks in https://github.com/jetstack/cert-manager/pull/888

PR #193, opened 2018-10-11
Prevent “|” from breaking tags-based searches

brandon-rhodes:
An interior “|” inside either the main pattern or the file path pattern
would cause the tags-based search to match many lines it should not,
since the subordinate patterns were being combined into a single pattern
without parens around the subordinate patterns.


nelhage:
Nice fix.

PR #195, opened 2018-11-22
Add configurable links to files and lines

daviddoran-stripe:
## Summary

This can be used to, for example, add links to open an editor such as "Open in VS Code" or web tools, etc.

Sample `livegrep.json` snippet:

```
"file_links": [
	{
		"label": "Open on GitHub (file only)",
		"url-pattern": "https://github.com/{name}/blob/master/src/{path}",
		"whitelist-repo-pattern": "^([^/]+/)?livegrep$"
	},
	{
		"label": "Open on GitHub (with line)",
		"url-pattern": "https://github.com/{name}/blob/master/src/{path}#L{lno}",
		"whitelist-repo-pattern": "^[^/]+/livegrep$"
	}		
]
```

## Test Plan

Check the following `livegrep.json` configurations:
- No `file_links` key works fine
- An empty `file_links` array works fine
- A link with empty `whitelist-repo-pattern` is shown for all repos
- Links are only shown for repos that match the regex in `whitelist-repo-pattern`

Check this functionality in the UI:
- Clicking a link in the file header uses the line number of the first match for `{lno}`
- Links are shown in each file result header
- Only links that contain `{lno}` are shown beside each matching line

## Screenshot
build search 2018-11-22 19-31-46

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -74,7 +74,7 @@ type RepoConfig struct {
 }
 
 type LinkConfig struct {
-	Label                string `json:"label"`
-	UrlPattern           string `json:"url-pattern"`
-	WhitelistRepoPattern string `json:"whitelist-repo-pattern"`
+	Label            string `json:"label"`
+	UrlTemplate      string `json:"url_template"`
+	WhitelistPattern string `json:"whitelist_pattern"`
 }
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -57,6 +57,8 @@ type Config struct {
 	IndexConfig IndexConfig `json:"index_config"`
 
 	DefaultSearchRepos []string `json:"default_search_repos"`
+
+	LinkConfigs []LinkConfig `json:"file_links"`
 }
 
 type IndexConfig struct {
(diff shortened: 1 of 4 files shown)


server/config/config.go
```diff
@@ -70,3 +72,9 @@ type RepoConfig struct {
 	Revisions []string          `json:"revisions"`
 	Metadata  map[string]string `json:"metadata"`
 }
+
+type LinkConfig struct {
+	Label                string `json:"label"`
+	UrlPattern           string `json:"url-pattern"`
```
nelhage: The inconsistent use of `-` and `_` bothers me here. Can we be consistent?

server/server.go
```diff
@@ -98,11 +98,21 @@ func (s *server) ServeSearch(ctx context.Context, w http.ResponseWriter, r *http
 		bk.I.Unlock()
 	}
 
+	link_configs := make([]map[string]string, len(s.config.LinkConfigs))
+	for idx, link_config := range s.config.LinkConfigs {
+		link_configs[idx] = map[string]string{
```
nelhage: Can we just pass the `s.config.LinkConfigs` object directly instead of destructuring it and restructuring it into an identically-shaped object with different names? If we switch to using `_` in the JSON it'll even still be easy-to-access as a Javascript object.

Issue #120, opened 2017-10-18
Unable to install node dependencies using bazel


badmon:
I get this error:
ERROR: /Users/raymond/code/livegrep/web/npm/css-loader/BUILD:7:1: installing node modules from web/npm/css-loader/npm-shrinkwrap.json failed (Exit 1).
/private/var/tmp/_bazel_raymond/b8234531df81349f097554a459f6baeb/bazel-sandbox/8119839809326862884/execroot/com_github_livegrep_livegrep/bazel-out/host/bin/external/org_dropbox_rules_node/node/tools/npm/install.runfiles/nodejs/bin/node: /private/var/tmp/_bazel_raymond/b8234531df81349f097554a459f6baeb/bazel-sandbox/8119839809326862884/execroot/com_github_livegrep_livegrep/bazel-out/host/bin/external/org_dropbox_rules_node/node/tools/npm/install.runfiles/nodejs/bin/node: cannot execute binary file
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_raymond/b8234531df81349f097554a459f6baeb/bazel-sandbox/8119839809326862884/execroot/com_github_livegrep_livegrep/bazel-out/host/bin/external/org_dropbox_rules_node/node/tools/npm/install.runfiles/com_github_livegrep_livegrep/../org_dropbox_rules_node/node/tools/npm/install.py", line 49, in <module>
    main()
  File "/private/var/tmp/_bazel_raymond/b8234531df81349f097554a459f6baeb/bazel-sandbox/8119839809326862884/execroot/com_github_livegrep_livegrep/bazel-out/host/bin/external/org_dropbox_rules_node/node/tools/npm/install.runfiles/com_github_livegrep_livegrep/../org_dropbox_rules_node/node/tools/npm/install.py", line 45, in main
    npm_install(args.shrinkwrap, args.output)
  File "/private/var/tmp/_bazel_raymond/b8234531df81349f097554a459f6baeb/bazel-sandbox/8119839809326862884/execroot/com_github_livegrep_livegrep/bazel-out/host/bin/external/org_dropbox_rules_node/node/tools/npm/install.runfiles/com_github_livegrep_livegrep/../org_dropbox_rules_node/node/tools/npm/install.py", line 27, in npm_install
    run_npm(['install'], env=env, cwd=output)
[...]

PR #204, opened 2019-01-21
Bump deps to build with Bazel 0.21

miikka:
Hi, I'm trying to try out livegrep! To build livegrep with Bazel 0.21 on macOS, I had to bump a number of rule dependencies – see the attached patch. In addition to that, I had to use the workaround in https://github.com/livegrep/livegrep/issues/120#issuecomment-337635796 and run Bazel with `--incompatible_package_name_is_a_function=false` due to use of deprecated variables by rules_node (patch at https://github.com/dropbox/rules_node/pull/14).

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -3,7 +3,6 @@ load("@compdb//:aspects.bzl", "compilation_database")
 
 compilation_database(
     name = "compilation_db",
-    exec_root_marker = True,
     targets = [
         "//src/tools:codesearch",
         "//src/tools:codesearchtool",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -43,7 +43,7 @@ git_repository(
 
 git_repository(
     name = "com_github_nelhage_rules_boost",
-    commit = "96ba810e48f4a28b85ee9c922f0b375274a97f98",
+    commit = "c1d618315fa152958baef8ea0d77043eebf7f573",
     remote = "https://github.com/nelhage/rules_boost",
 )
 # local_repository(
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -148,6 +148,6 @@ new_git_repository(
 package(default_visibility = ["//visibility:public"])
 """
     ),
-    commit = "02c33ed2c0e86053073080fd215f44356ef5b543",
+    commit = "7bc80f9355b09466fffabce24d463d65e37fcc0f",
     remote = "https://github.com/grailbio/bazel-compilation-database.git",
 )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,7 +37,7 @@ git_repository(
 
 git_repository(
     name = "gflags",
-    commit = "660603a3df1c400437260b51c55490a046a12e8a",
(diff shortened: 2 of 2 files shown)


miikka:
I'm not sure what to make of the CircleCI failure, being a complete Bazel newbie. :|

alexdebrie:
@miikka Thanks for opening this! I was working on it over the weekend and had some similar changes.

Two additional questions that I had to do to get it to work on my Mac:

1) For the `org_dropbox_rules_node`, did you need to change the `WORKSPACE` to allow for an OSX build [as stated here](https://github.com/dropbox/rules_node#macos)? If I didn't do that, I had problems executing `npm install` during the build as the binaries were built for Linux.

2) Did you have any trouble w/ the `sha256sum` command [here](https://github.com/livegrep/livegrep/blob/5aacba4bce494ad7ca5b07bb25e4a140c1731f87/web/BUILD#L40) ? I added `sha256sum` to my machine, but it looks like the command to execute it sets the PATH to `PATH=/bin:/usr/bin`, while mine is located as `/usr/local/bin/sha256sum`. OSX is pretty finicky about adding to `/usr/bin` -- did you just override it or am I missing something?

miikka:
1. You're right, I forgot that.
2. It can be solved by `--action_env=PATH`, which instructs Bazel to take `PATH` from your environment.

It seems that I also needed to use `--incompatible_remove_native_http_archive=false` due to use of `new_http_archive` in `org_pubref_protobuf`. Thus my build command was:

```sh
bazel build //... \
  --incompatible_package_name_is_a_function=false \
  --incompatible_remove_native_http_archive=false \
  --action_env=PATH
```

alexdebrie:
Perfect! Thanks, @miikka. 

I don't know much about Bazel and I know even less about C++, but I'll see if I can figure out that Circle build issue over the weekend.

alper:
Bazel tells me: `ERROR: Unrecognized option: --incompatible_package_name_is_a_function=false`

alexdebrie:
@alper What version of Bazel are you using (`bazel version`) ?

I think this flag was added in 0.13, [as that is when `PACKAGE_NAME` was deprecated](https://blog.bazel.build/2018/05/02/bazel-0.13.html).

miikka:
Bazel 0.22 removed that flag again, but after PRs #210 and #211 it or `--incompatible_remove_native_http_archive=false` should not be needed. Still need the [rules_node workaround for macOS](https://github.com/dropbox/rules_node#macos). My build command:
```
bazel build //... --action_env=PATH
```

alper:
That command still does not work for me. I'm using the Bazel that comes with homebrew.

miikka:
Well, I guess you're on your own with debugging, then. Something I'd try is to use [the version from Bazel's official tap](https://docs.bazel.build/versions/master/install-os-x.html#install-on-mac-os-x-homebrew) instead of the one from homebrew core. The official version uses a bundled JDK instead of the system one – using the system one may cause some extra problems.

alper:
That is literally what I did. I just followed the official instructions. I also would be surprised if I did have a system JDK because I didn't install one.

PR #205, opened 2019-01-25
Update CircleCI builds

alexdebrie:
Companion to #204. cc @miikka

This updates the Docker image used by CircleCI to use a newer version
of Bazel (0.21.0). It also adds a few flags to the bazel commands in
the CircleCI build script to account for deprecated usage in some
dependencies.

This build itself won't be using the new image. We'll need to build and push the new image, then update the tag in the `.circleci/config.yml` file.


nelhage:
Thanks! I'll grab this and build a new base image.

PR #207, opened 2019-01-26
Bump the base image

nelhage:
Bump the base image to pick up #205


nelhage:
cc @miikka @alexdebrie this folds in #204 and some further changes, so we now build with Bazel 0.21 in CI (and should work locally, as well)

nelhage:
Thanks for both your contributions to get this working!

alexdebrie:
Awesome, @nelhage . Thank you for tweaking and merging!

PR #210, opened 2019-01-30
Upgrade org_dropbox_rules to work with Bazel 0.22

miikka:
Hello again! Bazel 0.22 was released, it removed `--incompatible_package_name_is_a_function=false`, and the build is broken again. Luckily [rules_node](https://github.com/dropbox/rules_node) got some updates, so the flags in `.bazelrc` are not needed anymore.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,7 +1,0 @@
-# These are needed by rules_node
-build --incompatible_package_name_is_a_function=false
-build --incompatible_remove_native_http_archive=false
-query --incompatible_package_name_is_a_function=false
-query --incompatible_remove_native_http_archive=false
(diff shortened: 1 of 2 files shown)


nelhage:
Looks like someone still needs `incompatible_remove_native_http_archive`. Is that also gone in 0.22? If so, we can probably just bump other deps to get a fix, since it looks like it's in protobuf stuff…

miikka:
Ah, I didn't realize I need to use `--expunge` with `bazel clean` to properly test this. `incompatible_remove_native_http_archive` is still there, so I added the flag back.

nelhage:
Great, thanks! I'll try to find time to clean up the other one as well, since I'm sure it's on the way out.

PR #215, opened 2019-02-20
Allow GitHub-style line range anchors

lopopolo-stripe:
When linking to a range in GitHub, the syntax is #L43-L70. With livegrep,
the syntax is #L43-70. This inconsistency causes me to make a lot of
mistakes.

This patch makes the fileview frontend accept both anchor styles for
linking to line ranges.

cc @nelhage


nelhage:
Cool. I wonder if we should switch to generating this style, too, for consistency. I don't have a particularly strong opinion either way.

Issue #216, opened 2019-02-26
Index git submodules as well


staktrace:
If a git repo has submodules, those are currently skipped during the repo walk. There could be an option in the config to also process submodules.

If this is desirable I can try to write a patch.

nelhage:
I'd happily accept a patch. I'm a bit undecied about whether I think it makes sense for `codesearch` to do this itself (neither of the options of treating files in submodules as children of the parent repository, or of inventing a new `indexed_tree`  per submodule feels *quite* right), or for the `livegrep-*` indexing helpers to interpret submodules and add new entries to the config JSON. Which one are you envisioning for your use case?

staktrace:
I admit I'm not too familiar with the architecture, but based on my 3-minute glance through the code I was imagining that as `git_indexer` walks the tree, if it finds a submodule, it would just create a new `git_indexer` instance for it and recurse on it. That sounds like your option "inventing a new `indexed_tree` per submodule". I had additionally envisioned allowing users to opt-in to walking submodules via a flag in the config file, to preserve backwards compatibility.

For my use case I'd ideally want to just set a flag in the config JSON file that I pass to `codesearch` and have it automatically index the submodules.

staktrace:
Oh I guess I'd also want search results that refer to submodule files to have their path adjusted appropriately, so I don't have to do it on my end.

nelhage:
What do you want the paths adjusted to? The path from the root of the outer repository, or the path inside the submodule repository?

I think this gets to why I'm a bit uncertain about recursing into a new `git_indexer`; there's metadata (such as how paths should be handled, and what github repository or remote URL viewer) we store on indexed repositories that it's not easy to specify for submodules. But maybe the answer is, if you want any of that, it's easy enough to what https://github.com/staktrace/mozsearch/commit/c69cebe434226f171fea87084926c917daf26c21 does and reify them yourselves; the convenience behavior should just be, essentially, "treat this submodule as though it's just an in-place subtree". I'm open to something like that.

staktrace:
The path to the root of the outer repository is what I'd like. So yeah, treating it as an in-place subtree is effectively what I want. I'll see if I can write a patch that does it. Thanks for the quick feedback!

staktrace:
Sorry for the long delay here, I got busy with other stuff. Finally got around to testing the patch and it seems to work as intended. PR is #225

PR #225, opened 2019-04-18
Add a flag to include submodules when walking the tree.

staktrace:
The flag, when set to true, will treat submodules as though they
are just regular files inside the parent repo. The submodule is
indexed at the revision that the parent repo points to.

Fixes livegrep/livegrep#216


nelhage:
This looks great! Thanks!

PR #227, opened 2019-04-21
Pass the github key to the git fetch or clone, if given

### LINKED ISSUE #223: [Livegrep Github Reindex] Git Clone over https not working for private repositories
## What? 

When running `livegrep-github-reindex` script using `-github-token` and `-http` flag: 
* it is fetching a list of repositories as expected.
* but when is trying to clone it, asks for username/password. 

## Why?  

There is no implementation of tokens on clone code neither a GH(github) client injection with oauth http client.

nelhage:
fixes #223

### DIFF
--- a/cmd/livegrep-github-reindex/main.go
+++ b/cmd/livegrep-github-reindex/main.go
@@ -379,12 +379,47 @@ func checkoutWorker(dir string,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


nelhage:
@markomafs can you confirm that this works for you? I tested it locally and it seems to work here.

PR #228, opened 2019-05-06
Avoid undefined behavior in file_contents_builder::build.

benjaminp:
file_contents_builder::build was unconditionally constructing a file_contents object over the result of the allocator. If the allocator returned nullptr, undefined behavior happened. Move the nullptr check earlier to fix this.


nelhage:
Thanks! Was this caught by inspection, ASAN, or something else, out of curiosity?

benjaminp:
I was inspecting this function debugging a different problem and noticed it.

PR #231, opened 2019-05-17
Bump rules_proto and bazel-compilation-database for bazel compatibility

ekinhbayar:
Hi! Without these bumps I wasn't able to build livegrep with bazel 0.25.1, so here's a PR.


nelhage:
Thanks! I merged this via #232 to keep CI green.

PR #235, opened 2019-05-27
Add GitLab reindexing support

craigfurman:
Livegrep is a great tool, and it would be very useful to me on a daily basis if automatic reindexing supported GitLab as well as GitHub.

The `livegrep-github-reindex` binary is renamed to `livegrep-codehost-reindex`, and GitLab support is added. I found that most of the code in that module was reusable, so I extracted the repository list-building logic behind an interface, and implemented that interface again for GitLab.

Currently only only GitLab groups can be specified for indexing, but it wouldn't be difficult to add support for users and individual repos as is supported for GitHub. Projects inside subgroups are included, recursively.

The GitHub-specific flags are namespaced, e.g. `org` -> `github-org`.

Is this something you'd be interested in? I appreciate that renaming binaries and flags in a breaking way could be a little controversial!


nelhage:
Thanks for the PR! I'd definitely be interested in something of this shape.

I was actually thinking of redoing `livegrep-github-config` to accept more flexibility in its configuration anyways, and adding support for multiple hosting sites makes a lot of sense as another feature. Ideally I'd also integrate it into `livegrep-fetch-reindex`, too, I think. I'll take a look at this PR this weekend or so, hopefully. If I have time I might borrow your gitlab code but otherwise refactor into a new interface I have in my head, but we'll see.

craigfurman:
Seems reasonable! I'd be interested in seeing what becomes of this refactoring. Until then I can build the git{lab,hub} reindexer from my fork and use it with the rest of livegrep. Thanks again for a great project :+1:

thairu:
+1 to this issue :) I recently had to modify this to work with Bitbucket, and would be great to have the general solution mentioned in https://github.com/livegrep/livegrep/pull/235#issuecomment-497017953 

I solved it on my end by writing a separate tool to generate the config from Bitbucket, and used `livegrep-fetch-reindex` to handle the re-index using the config. It might make sense to have a separate tool for config generation, with site specific clients, and pipe the output of that to the existing `livegrep-fetch-reindex`, which only requires `git` access

NightMachinary:
@nelhage Any updates on Gitlab support?

nelhage:
Hey, I'm sorry for the radio silence for so long. I just pushed #276 which is basically an implementation of @thairu's suggestion. I think my preferred path forward here would be parallel `livegrep-bitbucket-reindex`/`livegrep-gitlab-reindex` commands, which chain to `livegrep-fetch-reindex`, which I've beefed up a bit with some of the enhancements that were previously only in the github fetcher. Does that approach make sense to @NightMachinary and/or @craigfurman? We could potentially also go in the direction of a "contrib" repository where we can add various 3rd-party reindexers that all back to to the common configuration format.

NightMachinary:
[**@nelhage**](https://github.com/nelhage) commented on [Aug 24, 2020, 11:12 PM GMT+4:30](https://github.com/livegrep/livegrep/pull/235#issuecomment-679299012 "2020-08-24T18:42:51Z - Replied by Github Reply Comments"):
> Hey, I'm sorry for the radio silence for so long. I just pushed [#276](https://github.com/livegrep/livegrep/pull/276) which is basically an implementation of [@thairu](https://github.com/thairu)'s suggestion. I think my preferred path forward here would be parallel `livegrep-bitbucket-reindex`/`livegrep-gitlab-reindex` commands, which chain to `livegrep-fetch-reindex`, which I've beefed up a bit with some of the enhancements that were previously only in the github fetcher. Does that approach make sense to [@NightMachinary](https://github.com/NightMachinary) and/or [@craigfurman](https://github.com/craigfurman)? We could potentially also go in the direction of a "contrib" repository where we can add various 3rd-party reindexers that all back to to the common configuration format.

I don't quite understand the architecture here to comment on this design decision, but it seems good enough.

PS: I decided to use Sourcegraph instead. Then I decided to use plain old CLI fzf+ripgrep.

craigfurman:
I've lost all context on this but that suggestion sounds plausible. Thanks @nelhage! I'll close this.

PR #244, opened 2019-11-19
Include unique index identity in search result response

LINKIWI:
This change proposes the inclusion of two new fields, `index_name` and `index_time` (the config-specified name of the index and the current index timestamp, respectively) in the `CodeSearchResult` message. Both of these field values are obtained from the existing server implementation's member variables.

This ties a unique index identity with every search result (i.e., a search request + one state of the index can deterministically be transformed into a search result). In our deployment of livegrep, the index file is rolled several times per day, so the same search request can yield a different set of results throughout the day.

Our livegrep client leverages a distributed caching layer that caches search results (keyed by a serialization of the query), and including a unique index identity in the search response allows for a more robust cache key design. By encoding the index name and timestamp in the key, whenever a new index is loaded by the server, all existing cache keys will be immediately invalid (since the index timestamp has now changed).

Let me know what you think--feel free to close if you feel this use case is too niche.

### DIFF
--- a/src/proto/livegrep.proto
+++ b/src/proto/livegrep.proto
@@ -70,6 +70,9 @@ message CodeSearchResult {
     SearchStats stats = 1;
     repeated SearchResult results = 2;
     repeated FileResult file_results = 3;
+    // unique index identity that served this request
+    string index_name = 4;
+    int64 index_time = 5;
 }
 
 message InfoRequest {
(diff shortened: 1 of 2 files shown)


nelhage:
This makes perfect sense to me! Purely out of curiosity, where are you using livegrep? Is your client open-source?

LINKIWI:
Our deployment is still experimental so I don't want to speak too soon :stuck_out_tongue:

The client is open source: https://github.com/LINKIWI/webgrep It is similar to the first-party client but has some other features to better fit our use cases.

PR #250, opened 2020-02-01
End-to-end server-side search latency stat

LINKIWI:
I've noticed that multithreaded searches (`--threads` param > 0) over an index with many chunks (built with a small-ish `--chunk_power`) result in inaccurate timings in the returned stats message. I'm guessing this is because the timers are run multiple times, once in each thread, causing the total time to be the sum of all individual latencies. This behavior seems correct from a literal standpoint of what the timers are measuring but also a bit unintuitive.

This change proposes the addition of a field to the stats message for an end-to-end "total" time taken by the search. It is measured at a reasonably high level (implementation of the `search` RPC method) in order to most faithfully capture the total server-side search latency.

### DIFF
--- a/src/proto/livegrep.proto
+++ b/src/proto/livegrep.proto
@@ -45,6 +45,7 @@ message SearchStats {
     int64 sort_time = 3;
     int64 index_time = 4;
     int64 analyze_time = 5;
+    int64 total_time = 7;
     enum ExitReason {
         NONE = 0;
         TIMEOUT = 1;
--- a/src/tools/grpc_server.cc
+++ b/src/tools/grpc_server.cc
@@ -337,6 +337,7 @@ Status CodeSearchImpl::Search(ServerContext* context, const ::Query* request, ::
         ;
 
     match_stats stats;
+    timer search_tm(true);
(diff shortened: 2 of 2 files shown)


LINKIWI:
For example:

```
$ time grpc_cli call localhost:20027 CodeSearch.Search 'line: "test"' | grep --context 5 stats
--call_creds=none invalid; must be none or access_token=<token>.
connecting to localhost:20027
Rpc succeeded with OK status
stats {
  re2_time: 12
  git_time: 74
  exit_reason: MATCH_LIMIT
  total_time: 37
}

real	0m0.047s
user	0m0.004s
sys	0m0.000s
```

nelhage:
This makes sense to me. I wonder if we can do better than "total_time" for the name, but I don't have a suggestion I like much more. I think I'm happy to take as-is.

LINKIWI:
One alternative I had in mind was `e2e_time` but I figured it may be confusing since it only covers latency of the search logic routine (excluding e.g. RPC mechanics, network I/O, etc.)

jboning:
Driveby idea, perhaps something involving "span"? neither `span_time` or `time_span` really ring to me, but the meaning is in the right ballpark.

PR #251, opened 2020-02-25
Interface to bypass server-side match limiter per request

LINKIWI:
This offers a small change to the client API that allows clients to bypass the server-side match-based search limiter.

Proposed behavior: When `max_matches` is specified as 0, we will defer to `FLAGS_max_matches`, which in turn defaults to 50 when not specified. When `max_matches` is specified as any negative integer, we explicitly disable the match limiter (the search itself is still subject to the timeout limiter, if enabled).

This is a deviation from current behavior where a zero or negative value of `max_matches` always defers to `FLAGS_max_matches`.

I believe this PR is the only change needed, since the following logic will ignore the match limit if it is zero-valued:

https://github.com/livegrep/livegrep/blob/beca552bc53f644920d78e64504b2c5a84d952a9/src/codesearch.cc#L140

### DIFF
--- a/src/tools/grpc_server.cc
+++ b/src/tools/grpc_server.cc
@@ -293,8 +293,13 @@ Status CodeSearchImpl::Search(ServerContext* context, const ::Query* request, ::
     q.trace_id = current_trace_id();
 
     q.max_matches = request->max_matches();
-    if (q.max_matches <= 0 && FLAGS_max_matches)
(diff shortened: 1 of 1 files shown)


nelhage:
This makes sense to me! As best I can tell you can already _de facto_ do this by passing an ~arbitrarily-high `max_matches` so I don't think this is that much of a functional change, just a sanity improvement.

PR #265, opened 2020-04-30
Skip archived repositories when indexing.

### LINKED ISSUE #264: reindex: option to ignore archived repos
There's an option to blacklist certain repos in an org:
https://github.com/livegrep/livegrep/blob/3bcd52d4531e61505117aa251265fc93de26d7b1/cmd/livegrep-github-reindex/main.go#L29

We are archiving a bunch of repos and don't want those results to appear, but find ourselves having to maintain a mirror of that info in our livegrep config.

nelhage:
`-archived=true` is available if anyone needs the old behavior.

fixes #264


trevorcreech:
Amazing, I just came here to make a PR to do this :D

PR #273, opened 2020-08-01
Add jump to definition and hover using LSPs

brian-yu:
This PR builds off of the work done @ilinum and other Dropbox folks in #174 as well as work done by @ains. 

In addition to resolving merge conflicts in #174, @dmssargent, @yhylord, and I added support for LSP's hover feature.

Examples:
image

image

We've tested it with [`javascript-typescript-langserver`](https://github.com/sourcegraph/javascript-typescript-langserver), [`go-langserver`](https://github.com/sourcegraph/go-langserver) and [`sorbet`](https://github.com/sorbet/sorbet).

To try it, run `go-langserver -addr ":8040" -mode "tcp"` and configure the `lang_servers` option in `server.json` like so:

```json
{
    "name": "livegrep",
    "repositories": [
        {
            "name": "livegrep",
            "path": "/livegrep",
            "revisions": [ "HEAD" ],
            "metadata": {
                "github": "nelhage/livegrep",
                "url-pattern": "https://github.com/{name}/blob/HEAD/src/{path}#L{lno}",
                "blame": "git"
            },
            "lang_servers": [
                {
                    "address": "0.0.0.0:1234",
                    "extensions": [
                        ".go"
                    ]
                }
            ]
        }
    ]
}
```


yhylord:
Update since last comment: Since we don't know how to highlight the code inside hovering tooltip, we removed lines starting with ``` to make it look less scary. We would like to know how to highlight it and revert this change however.

One more note: We have implemented jumping to references on the backend, but we don't know how to present 100+ references on the frontend. Would like some tip on this.

nelhage:
Hey! Sorry for the slow response, I've been on vacation a while and then slow to catch up. This is exciting!

I'm going to take a quick look at the code, but first I wanted to ask some questions about whether this is deployed and whether there is appetite and energy to contribute to maintaining this, either at Stripe, Dropbox, or elsewhere. I probably won't deploy this infrastructure anywhere myself, and so I'd be much happier eventually merging it if it's actually in use and if someone else will be motivated to handle bugfixes and other maintenance as it comes up.

.gitignore
```diff
@@ -16,3 +16,5 @@ vendor/re2/obj/libre2.a
 /bazel-*
 compile_commands.json
 .devbox
+.DS_Store
```
nelhage, line 19: I'm happy to take this change, but IMO this is the kind of thing that better lives in a [`core.excludesFile`](https://git-scm.com/docs/gitignore#_configuration) so that it is globally excluded on a developer's machine, instead of on a per-project basis. But since that feature is admittedly a bit esoteric, I'm happy to add it to the project so things "just work" a bit better for contributors on macs.

web/src/fileview/fileview.js
```diff
@@ -195,6 +195,162 @@ function init(initData) {
...
+
+  // When source code is hovered over, highlight/underline any tokens for which
+  // jump-to-definition will work.
+  function onHover(clientX, clientY) {
+    // The source-code consists of a <code id='source-code' class='code-pane'>
+    // containing lots of <span class="token tokentype">token</span>
+    // hoverable text may be surrounded by <span class="hoverable">
+    // text being hovered over is changed to class="hovering"
+    // non-hoverable text is class="nonhoverable"
+    const pos = document.caretRangeFromPoint(clientX, clientY);
```
nelhage, line 315: It looks like [caretRangeFromPoint](https://developer.mozilla.org/en-US/docs/Web/API/Document/caretRangeFromPoint) is nonstandard and doesn't exist in Firefox. Could we wrap this with a wrapper that uses either `caretRangeFromPoint` or [`caretPositionFromPoint`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/caretPositionFromPoint) (which is standards-track but experimental and only present in Firefox at present, AFAICT), depending on which one is present?

web/src/fileview/fileview.js
```diff
@@ -195,6 +195,162 @@ function init(initData) {
...
+    const textNode = pos.startContainer;
+    if (!textNode || textNode.nodeType !== 3) { // expected to be a text node
```
nelhage, line 317:
```suggestion
    if (!textNode || textNode.nodeType !== Node.TEXT_NODE) {
```

server/server.go
```diff
@@ -281,6 +299,136 @@ func (h *reloadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		if repo.Name == repoName {
+			return &repo, nil
+		}
+	}
+	return nil, errors.New(fmt.Sprintf("repo with name %s not found", repoName))
+}
+
+func (s *server) parseDocPositionParams(params url.Values) (*langserver.TextDocumentPositionParams, *config.RepoConfig, error) {
+	row, column, repo, fpath :=
+		params.Get(rowParamName),
```
nelhage, line 391: We use `row` in the code, but LSP uses `line` -- is that a deliberate decision? I suspect it would be less confusing to standardize on `line`, or maybe `line_no` if we want to imply that it's a line number and not a string line of text.

server/server.go
```diff
@@ -363,6 +514,31 @@ func New(cfg *config.Config) (http.Handler, error) {
...
+			client, err := langserver.NewClient(langServer.Address, &langserver.InitializeParams{
+				ProcessId:        nil,
+				OriginalRootPath: r.Path,
+				RootUri:          "file://" + r.Path,
+				Capabilities:     langserver.ClientCapabilities{},
+			})
+
+			// a failing language server isn't fatal. We'd prefer to log these metrics.
+			if err != nil {
+				log.Printf(ctx, "%s", err.Error())
```
nelhage, line 531: Let's provide some more context to this error, something like 
```go
log.Printf(ctx, "Failed to create LSP client: %s@%s: %s", r.Name, langServer.Address, err.Error())
```

brian-yu:
Hi! No worries at all, I hope you had a great vacation!

@dmssargent, @yhylord, and I have wrapped up our internships at Stripe (for reference, this was a hackathon project that we were able to deploy on QA before we left). Maybe @ains-stripe or @areitz would have some thoughts on the future of this at Stripe and perhaps @ilinum knows about the status of this at Dropbox.

Regardless, there's no rush to merge this PR and @dmssargent, @yhylord, and I are happy to help in any way we can in the meantime.

server/server.go
```diff
@@ -350,6 +499,8 @@ func New(cfg *config.Config) (http.Handler, error) {
 
 	m.Add("GET", "/api/v1/search/:backend", srv.Handler(srv.ServeAPISearch))
 	m.Add("GET", "/api/v1/search/", srv.Handler(srv.ServeAPISearch))
+	m.Add("GET", "/api/v1/langserver/jumptodef", srv.Handler(srv.ServeJumpToDef))
```
nelhage, line 502: I don't love `langserver` as the path component here. I'd like either `lsp`, which is the standard abbreviation for the language server protocol, or perhaps something like `sema` or `analyze` which gets at the *logical* function of these endpoints -- these are the endpoints that provide semantic source-code analysis. It happens to be backed by LSP today, but in principle you could imagine swapping out the server implementation for something else, while preserving at least approximately the same server API…

How do you feel about using `lsp` instead of `langserver` here? There's a perfectly good standard acronym for what we're using

.gitignore
brian-yu: TIL! I'll remove these changes in favor of `core.excludesFile`.

web/src/fileview/fileview.js
brian-yu: Good catch, done.

server/server.go
brian-yu: Done. Also switched to `character` instead of `col` to be more consistent with LSP.

server/server.go
brian-yu: I agree that`lsp` is better. Fixed 😄

PR #274, opened 2020-08-13
inotify-based hot index file reloads

LINKIWI:
This change proposes a new flag `--hot_index_reload` which uses a similar mechanism as `--reload_rpc` to automatically reload the server when the index file on disk changes, using inotify C APIs. This behavior only applies when the server is started with a prebuilt index file (i.e. `--load_index`) and is incompatible with `--reload_rpc` (so that the shutdown thread only waits on a single blocking operations for restarts). As implemented, the server is restarted by touching, copying, or moving a new index to the path specified by `--load_index`.

Our indexing pipeline builds and deploys a new index file every few hours. This change allows the server to automatically pick up the refreshed index without needing to configure a client to issue a manual reload after the deploy completes.

Unfortunately, this only works on Linux due to the use of inotify. Feel free to reject this change if there's a requirement for livegrep to compile cross-platform.

### DIFF
--- a/src/lib/fs_linux.cc
+++ b/src/lib/fs_linux.cc
@@ -24,9 +24,6 @@ fswatcher::fswatcher(const std::string &path) : path_(path) {
 fswatcher::~fswatcher() {
     if (fd >= 0) {
         close(fd);
-        if (wd >= 0) {
-            inotify_rm_watch(fd, wd);
-        }
     }
 }
 
--- a/src/lib/fs_linux.cc
+++ b/src/lib/fs_linux.cc
@@ -17,22 +17,29 @@ namespace {
 }
 
 fswatcher::fswatcher(const std::string &path) : path_(path) {
-    fd = inotify_init();
-    wd = inotify_add_watch(fd, path.c_str(), IN_ATTRIB | IN_CLOSE_WRITE | IN_MOVE_SELF);
+    if ((fd = inotify_init()) > 0) {
+        wd = inotify_add_watch(fd, path.c_str(), IN_ATTRIB | IN_CLOSE_WRITE | IN_MOVE_SELF);
+    }
 }
 
 fswatcher::~fswatcher() {
-    if (fd >= 0) {
+    if (fd != -1) {
         close(fd);
     }
 }
 
-void fswatcher::wait_for_event() {
+bool fswatcher::wait_for_event() {
     struct inotify_event event;
     int n = 0;
+
+    if (fd == -1 || wd == -1) {
+        return false;
+    }
 
     // The read syscall is blocking; it returns after one eligible event (i.e., matching the mask) is received.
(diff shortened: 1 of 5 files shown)


nelhage:
I'm open to this change, but I would like to preserve our ability to compile outside of Linux -- In particular, I've wanted to do development from macOS in the past. I would find it acceptable for this feature to just error out on non-Linux builds, or to fall back on polling `stat` every N seconds. If you can make one of those changes I'll take a look!

LINKIWI:
Thanks, I'll investigate how to make it compile and error out gracefully at runtime on Mac. Unfortunately I don't have a Mac environment handy so it may take some time before I have something.

LINKIWI:
I've chosen to add an additional layer of abstraction: `src/lib/fs.h` for the interface with `src/lib/fs_linux.cc` and `src/lib/fs_default.cc` implementations, chosen by Bazel based on the host platform.

Right now, the default (non-Linux) implementation is a noop and prints some text to stderr when `--hot_index_reload` is specified. Let me know what you think, thanks.

src/lib/fs_linux.cc
```diff
@@ -0,0 +1,41 @@
...
+    }
+}
+
+void fswatcher::wait_for_event() {
+    struct inotify_event event;
+    int n = 0;
+
+    // The read syscall is blocking; it returns after one eligible event (i.e., matching the mask) is received.
+    while (n <= 0) {
+        n = read(fd, &event, sizeof(struct inotify_event) + NAME_MAX + 1);
```
nelhage, line 39: In the event that inotify is unavailable or `inotify_init` otherwise fails, am I correct that this will always immediately exit, causing a spinloop infinitely reloading the index?

src/lib/fs_linux.cc
```diff
@@ -0,0 +1,41 @@
...
+fswatcher::fswatcher(const std::string &path) : path_(path) {
+    fd = inotify_init();
+    wd = inotify_add_watch(fd, path.c_str(), IN_ATTRIB | IN_CLOSE_WRITE | IN_MOVE_SELF);
+}
+
+fswatcher::~fswatcher() {
+    if (fd >= 0) {
+        close(fd);
+        if (wd >= 0) {
+            inotify_rm_watch(fd, wd);
```
nelhage, line 28: Removing a watch _after_ we close the `fd` doesn't seem right. I think we don't need to clean up the watch at all if we're closing the `inotify` fd entirely, if I read `inotify(7)` right.

src/lib/fs_linux.cc
LINKIWI: That is correct. Let me see if I can come up with a better way to surface `inotify_init` and `inotify_add_watch` errors, and propagate them to the callsite.

nelhage:
Looks great, thanks!

PR #280, opened 2020-09-27
Encode filename index in index file dump

### LINKED ISSUE #279: Possibility of encoding filename index directly in index dump file
In a deployment with tens of millions of lines of source code, I've noticed that the `index_filenames` step https://github.com/livegrep/livegrep/blob/7b3cccb9c9e6201fb309f72fc4d580807e65bd12/src/dump_load.cc#L496 invoked on startup with `--load_index` can take a ~couple minutes to complete, which causes the server to be unavailable during this short period of time while the index is reloaded.

I'm curious if it's feasible to precompute the filename index and encode it into the dump file so that it can be loaded more quickly on startup? Seems like it would a breaking change to the format specification but could yield performance improvements even if read into memory eagerly, i.e. without creating an mmap'd segment for it for lazy access.

I can try taking a stab at this myself, wanted to check if it's possible or even a worthwhile idea.

LINKIWI:
This change proposal attempts to fix #279. It includes the following high level changes:

- Expand the index header to include sizes and offsets of [all three data structures used to store the filename index](https://github.com/livegrep/livegrep/blob/7b3cccb9c9e6201fb309f72fc4d580807e65bd12/src/codesearch.h#L231)
- Encode said data structures into the dump in a new routine `dump_filename_index` in `codesearch_index`, and decode their contents into a `code_searcher` instance on a `load_index` operation
- Bump the index version constant due to a backwards-incompatibile index file format change
- Remove the `index_filenames()` step invoked when loading an offline-built index file

This seems to work in my very limited manual testing. For an index with ~75M lines of source code over ~300K files, this cuts the gRPC server launch time from ~4s to ~1s on my machine. (It's already quite fast as-is; the startup/reload slowdown mentioned in #279 ended up being due to an unrelated issue)

### DIFF
--- a/src/codesearch.cc
+++ b/src/codesearch.cc
@@ -505,13 +505,12 @@ void code_searcher::index_filenames() {
 void code_searcher::finalize() {
     assert(!finalized_);
     finalized_ = true;
+    index_filenames();
     alloc_->finalize();
 
     timeval now;
     gettimeofday(&now, NULL);
     index_timestamp_ = now.tv_sec;
-
-    index_filenames();
 
     idx_data_chunks.inc(alloc_->end() - alloc_->begin());
     idx_content_chunks.inc(alloc_->end_content() - alloc_->begin_content());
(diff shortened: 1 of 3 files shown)


nelhage:
Looks great, thanks for the PR!

PR #281, opened 2020-10-10
Avoid messy stderr output in some cases

staktrace:
If the repo being indexed has files that are too large, a WARN
line is emitted between the "walking HEAD..." and "done" outputs.
This results in messy stderr that looks like this:

  walking HEAD... WARN: file is too big to be indexed
WARN: file2 is too big to be indexed
done

Apart from looking messy, this makes it harder to autodetect
warnings with grep. It seems better to insert a newline
after the "walking HEAD..." output, which also makes it
consistent with the other branch of code immediately above.


nelhage:
Makes a lot of sense, thanks!

PR #286, opened 2021-03-15
Pin Bazel 4.0.0 in .bazelversion

LINKIWI:
This small change proposes the addition of a [`.bazelversion`](https://docs.bazel.build/versions/master/updating-bazel.html) file used by [bazelisk](https://github.com/bazelbuild/bazelisk) for building livegrep with a specific Bazel version.

With the recent forward compatibilty fixes, everything builds on 4.0.0.


nelhage:
Neat! I hadn't seen bazelisk, but that seems like a great approach.

Looks like I broke CI, but I'll force-merge this for now while I fix.

PR #294, opened 2021-04-30
fix bug where file extensions weren't counted for path only results

lunixbochs:
Let's say you do a search for "foo" and it results in:
```
src/path/file.foo
src/bar.cpp: "foo"
```

The current code will only add `.cpp` to the common extension list from `bar.cpp`, despite `file.foo` also showing up in the file path results.

This PR fixes that by considering file paths for the extension list.


nelhage (approved):
Thanks!

Issue #296, opened 2021-06-03
Compile errors for BigSur


ptarjan:
There is some sort of python2 vs python3 incompatibility. When building with `master` I get this on BigSur:

```
$ bazel build //...
DEBUG: Rule 'com_grail_bazel_compdb' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1571147747 -0700"
DEBUG: Repository com_grail_bazel_compdb instantiated at:
  /Users/paul.tarjan/robinhood/opensource/livegrep/WORKSPACE:124:15: in <toplevel>
Repository rule git_repository defined at:
  /private/var/tmp/_bazel_paul.tarjan/2a6c7e36012cc56ce6d4dcd08beefc0e/external/bazel_tools/tools/build_defs/repo/git.bzl:199:33: in <toplevel>
INFO: Analyzed 57 targets (0 packages loaded, 0 targets configured).
INFO: Found 57 targets...
ERROR: /Users/paul.tarjan/robinhood/opensource/livegrep/web/npm/style-loader/BUILD:7:12: installing node modules from web/npm/style-loader/npm-shrinkwrap.json failed: (Exit 1): install failed: error executing command bazel-out/host/bin/external/org_dropbox_rules_node/node/tools/npm/install web/npm/style-loader/npm-shrinkwrap.json bazel-out/darwin-fastbuild/bin/web/npm/style-loader

Use --sandbox_debug to see verbose messages from the sandbox install failed: error executing command bazel-out/host/bin/external/org_dropbox_rules_node/node/tools/npm/install web/npm/style-loader/npm-shrinkwrap.json bazel-out/darwin-fastbuild/bin/web/npm/style-loader

Use --sandbox_debug to see verbose messages from the sandbox
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_paul.tarjan/2a6c7e36012cc56ce6d4dcd08beefc0e/sandbox/darwin-sandbox/2568/execroot/com_github_livegrep_livegrep/bazel-out/host/bin/external/org_dropbox_rules_node/node/tools/npm/install.runfiles/org_dropbox_rules_node/node/tools/npm/utils.py", line 92, in run_npm
    ret = subprocess.check_output(
  File "/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
[...]

PR #299, opened 2021-06-11
Support nodejs on darwin-x64 and linux-x64

clint-stripe:
Quick and kinda dirty fix for #296.

The "error" appeared to be python related, but it was actually a failure from the python wrapper trying to run the linux binary on a Mac.

This could probably use some refinement, notably we don't _fail_ if you're on Windows, darwin-arm64, etc.


nelhage:
Ah-ha, good find!

PR #300, opened 2021-07-01
Add some more python-ish files to syntax highlighting

clint-stripe:
Adding `BUILD.bazel` and `.sky` (for Starlark files) to use python-ish highlighting too.

Thoughts on adding a config section so this doesn't need to be a code change?


nelhage:
Thanks! Yeah, I think adding a configuration option – ideally, while retaining some defaults internally – would make sense here.

PR #303, opened 2021-07-16
Add a max-width span around file name link

danwang:
On the search results page, one thing I've noticed is that copying the path from the header results in a `^M` newline character being copied as well.

One way to reproduce this is to start a search, then click and drag your mouse starting where the red arrow is (e.g., to select the text `web/src/codesearch/codesearch_ui.js`)

Screen Shot 2021-07-16 at 3 39 37 PM

When you paste, it also pastes a newline. This adds another step to delete the newline if you're opening a file by pasting the path into something like vim's `:e` command.

To fix, I wrapped the `<a>` in big span. It's bigger than the link, so only the text remains the target, but you can still highlight and not get the newline. No visual differences detected.

Before:
Screen Shot 2021-07-16 at 3 46 39 PM

After:
Screen Shot 2021-07-16 at 3 46 16 PM

### DIFF
--- a/web/htdocs/assets/css/codesearch.css
+++ b/web/htdocs/assets/css/codesearch.css
@@ -216,6 +216,10 @@ a:hover {
     padding-right: 3px;
 }
 
+.header-path {
+    flex-grow: 1;
+}
+
 .result-path {
     color: #3d464d;
     font-family: "Menlo", "Consolas", "Monaco", monospace;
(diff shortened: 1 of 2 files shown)


nelhage:
Thanks! I _think_ that's a real CI failure from a javascript parse error in the pipeline, although I'm not 100% certain. Can you take a look?

danwang:
Tried pushing a fix. Can you help me kick off another CI run?

Issue #305, opened 2021-08-12
File names no longer showing up


bmalehorn:
We recently upgraded to the latest commit and found that our file names had vanished!

| before (ae31f02999) | after (186e6a6986) |
| --- | --- |
| ![image](https://user-images.githubusercontent.com/3344958/129124147-abd2bb84-a935-4281-adf1-f5332d7edf98.png) | ![image](https://user-images.githubusercontent.com/3344958/129124144-6298873d-13b0-4198-bbdd-808f238f2a08.png) |

This seems to be caused by https://github.com/livegrep/livegrep/pull/303, which changed the styling on file names. Rolling back to the previous commit, ae31f02999, fixed the issue.

@danwang would you mind taking a look at this?

danwang:
Sorry about that! I'll take a look at fixing this.

bmalehorn:
Thanks!

PR #306, opened 2021-08-13
Fix path names

danwang:
Fixes bug #305, which was introduced in #303. 

This time I got the right version of bazel and tested it locally all the way. Sorry. The second argument to `h.span` needs to be an array, even if there is only one element as children.


nelhage:
Thanks for looking into this!

Issue #313, opened 2021-12-14
Made some changes in fork, would any be useful? 


xvandish:
Hi, I have a few changes in my fork, and I guess before I go presumptively cherry-picking the one's that would be useful (if any) and opening a PR - I figured I'd ask if you'd like to see any of these.

1. Switched all CI to GH Actions and packaging system to GH Packages - would require some changes from `xvandish-livegrep` to `livegrep` as the package name. 
2. Allow num of pages per GH repos query (in github-live-reindexer) to be configurable (tiny change)
3. allow num of workers in github-fetch-reinfex to be configurable (also tiny change)
4. allow GH queries for repos to be concurrent at a configurable rate. For an organization with like 4k repos, saves a lot of time. ~50s 

Since some of these changes are in `main` already, but some aren't (namely concurrent requests) a PR would include the changes in main, so I'd have to cherry pick depending on what you would/wouldn't find useful.

Here's a comparison with the changes - let me know.
https://github.com/livegrep/livegrep/compare/main...xvandish:concurrent-reindexer-github-requests

I've also got a `statsd` implementation that I'll probably finish sometime soon, and I'm working on porting over/cleaning up the features that Dropbox references here (https://github.com/livegrep/livegrep/blob/d0148a0fe662d2fc10aebcbfcaac87cbeb8b17e6/web/templates/fileview.html#L85-L86) but that didn't actually get implemented (in the base repo, they are implemented in their fork) - if any of these future modifications are of interest, also let me know! Thanks.

nelhage:
Those all sound great! (1) especially is something I've been meaning to do.

Would you be open to splitting your fork into separate PRs for each feature? I'd be happy to review them in that form.

xvandish:
Cool, & ofc! I'll work on splitting out the feature commits onto new branches so they can by PR'd individually - & then open them up for your review.

xvandish:
PR #314 is up for change 2

xvandish:
PR #315 is up for changes 3 & 4.

xvandish:
these have all been made - closing

PR #314, opened 2021-12-19
bump PerPage value for gh-reindexer repo fetching to max

xvandish:
~~Small change to allow the number of repos per GitHub api request to be configurable. I mention the max is 100, but don't do any validation because the GH REST api does it. `go-github` doesn't validate this either, as far as I [can tell](https://github.com/google/go-github/blob/85425ec5f1e4118be4b578e5b6ad07f9ea32cf5f/github/repos.go#L283).  This is change 2 from issue #313~~

PR now simply bumps the PerPage number to `100`, which is the max allowed by the github REST api. Should have no effect on users/orgs that have less than 100  repos, while always reducing the num of network calls made for those that have >= 100 repos.

### DIFF
--- a/cmd/livegrep-github-reindex/main.go
+++ b/cmd/livegrep-github-reindex/main.go
@@ -328,7 +328,7 @@ func getOneRepo(client *github.Client, repo string) ([]*github.Repository, error
 func getOrgRepos(client *github.Client, org string) ([]*github.Repository, error) {
 	var buf []*github.Repository
 	opt := &github.RepositoryListByOrgOptions{
-		ListOptions: github.ListOptions{PerPage: 50},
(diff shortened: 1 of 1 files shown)


nelhage:
Is there a reason not to just hardcode this to the maximum? We're iterating over the entire list so I think that's just always the best choice.

xvandish:
> 



> Is there a reason not to just hardcode this to the maximum? We're iterating over the entire list so I think that's just always the best choice.

Good point! I just defaulted to trying to not change existing behavior - but agreed that it's probably always the best choice.  I'll update the default and the message lang.

xvandish:
@nelhage PR updated to simply hard code to max!

nelhage:
Thanks!

PR #315, opened 2021-12-19
Concurrent reindexer GitHub requests

xvandish:
This change does two things:
1. allows `livegrep-github-reindex` to make requests for repos (per user or per org) concurrently at a specified limit. By default it's set to 1, which retains the current behavior wherein only 1 call pero user/org is made at a time. Settings the limit to ~40 for me dropped the time the `updating repos` step took from ~1min to ~5s when indexing 2 orgs with a combined ~4k repos.
2. adds a new flag, `num-workers` to `livegrep-fetch-reindex` that updates the number of workers used to run the `checkoutRepos` function. Also updates `livegrep-github-reindex` and adds the arg `num-repo-update-workers` whose value will be bassed to the call to `livegrep-fetch-reindex` as `num-workers`

For the concurrent requests, I made the decision to try to retain order (the results of `repos?page=1` come before `repos?page=2` and so on), in case someone is expecting/relying on their repos to be updated/cloned/indexed in the order they're listed by the github api. Could remove some complexity if we decide that order isn't a big deal. Also, once things look god I can squash and clean up history a bit.

This is items 3 and 4 in #313 .

If any questions/comments at all let me know!

### DIFF
--- a/cmd/livegrep-fetch-reindex/main.go
+++ b/cmd/livegrep-fetch-reindex/main.go
@@ -24,9 +24,8 @@ var (
 	flagRevparse      = flag.Bool("revparse", true, "whether to `git rev-parse` the provided revision in generated links")
 	flagSkipMissing   = flag.Bool("skip-missing", false, "skip repositories where the specified revision is missing")
 	flagReloadBackend = flag.String("reload-backend", "", "Backend to send a Reload RPC to")
+	flagNumWorkers    = flag.Int("num-workers", 8, "Number of workers used to update repositories")
 )
-
-const Workers = 8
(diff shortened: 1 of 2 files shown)


nelhage:
Taking a look but it seems CI is failing for a real reason.

nelhage:
Looks good to me if you can get the build green!

xvandish:
Awesome, thanks for taking a look @nelhage ! Got the build to pass. Creating a feature branch out of an un-merged feature branch almost always comes back to haunt me yet I keep doing it. I'll learn someday.

PR #316, opened 2022-01-18
Add option for noindex to livegrep-github-reindex

juliannadeau-stripe:
This PR adds a flag to skip indexing in `cmd/livegrep-github-reindex/main.go`. At Stripe we use livegrep over a couple instances of GitHub, so our re-index operation calls this command a few times, combines the index, and generates the final index. This means that we are indexing `n + 1` times (where n is the number of GitHub locations) but only the last one matters.

This PR adds an easy option to simply return after indexing.


nelhage:
Thanks for the PR! This makes sense to me.

PR #322, opened 2022-02-18
Move core logic of no-index to fetch-reindex

juliannadeau-stripe:
The original PR implemted the logic in github-reindex not realizing that the fetching of the repos was tightly
coupled to fetch-reindex. This means that, with the old implementation, you could run the config generation but nothing
would actually update. You would just have HEAD but the repos would still be outdated.

Ideally we could de-couple github operations from indexing operations and independently fetch/index, but saving a full refactor
the easiest solution is to pass the flag into fetch-reindex and exit after fetching repos

### DIFF
--- a/cmd/livegrep-github-reindex/main.go
+++ b/cmd/livegrep-github-reindex/main.go
@@ -49,7 +49,7 @@ var (
 	flagDepth                   = flag.Int("depth", 0, "clone repository with specify --depth=N depth.")
 	flagSkipMissing             = flag.Bool("skip-missing", false, "skip repositories where the specified revision is missing")
 	flagMaxConcurrentGHRequests = flag.Int("max-concurrent-gh-requests", 1, "Applied per org/user. If fetching 2 orgs, you will have 2x{yourInput} network calls possible at a time")
-	flagNoIndex                 = flag.Bool("no-index", false, "Skip indexing after writing config")
+	flagNoIndex                 = flag.Bool("no-index", false, "Skip indexing after writing config and fetching")
 
 	flagRepos = stringList{}
 	flagOrgs  = stringList{}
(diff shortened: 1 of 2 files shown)


cmd/livegrep-github-reindex/main.go
```diff
@@ -140,17 +140,14 @@ func main() {
...
-		return
-	}
 
 	index := flagIndexPath.Get().(string)
 
 	args := []string{
 		"--out", index,
 		"--codesearch", *flagCodesearch,
 		"--num-workers", *flagNumRepoUpdateWorkers,
+		"--no-index", *flagNoIndex,
```
nelhage, line 150: You need to format this bool as a string here -- that's where the CI failure is coming from.

juliannadeau-stripe:
This was pretty late in my day heading into a long weekend, so I haven't tested quite yet. I opted to get a PR up and passing CI as I was having some Bazel trouble locally. Thanks for your patience here <3

cmd/livegrep-github-reindex/main.go
```diff
@@ -140,17 +141,14 @@ func main() {
...
-		return
-	}
 
 	index := flagIndexPath.Get().(string)
 
 	args := []string{
 		"--out", index,
 		"--codesearch", *flagCodesearch,
 		"--num-workers", *flagNumRepoUpdateWorkers,
+		"--no-index", strconv.FormatBool(*flagNoIndex),
```
vito-stripe, line 151: I think you'll need to conditionally append this instead like the other bool flags below or use `--no-index=...` to make it one arg. A lot of CLI parsers ignore separate bool flag args and treat the `false` as an arg, and commands that don't handle args often just drop them on the floor instead of asserting they're empty.

vito-stripe (changes requested):

vito-stripe:
Just tested this change in a deployment that initially had ~7 day stale checkouts, and everything worked as expected. Ran a re-index, saw the `--no-index` flag make it to the right place in `ps`, queried the API, and found the latest results.

Fixed the flag issue above too, so I think this is good to go!

nelhage:
Thanks, both of you!

PR #323, opened 2022-02-23
dump_load: Option for eager mmap prefault with MAP_POPULATE on Linux

LINKIWI:
[`mmap(2)`](https://linux.die.net/man/2/mmap) on Linux recognizes the `MAP_POPULATE` flag which eagerly prefaults the memory-mapped region into virtual memory, which reduces the rate of page misses during I/O on the memory-mapped file.

This change:

* Adds a flag `--eager_memory_load` to opt-in to eager prefaulting, supported only on Linux.
* Changes the existing `MAP_SHARED` flag to `MAP_PRIVATE`. Use of `MAP_POPULATE` requires `MAP_PRIVATE` on Linux. `MAP_SHARED` allows the memory-mapped region to be visible to other processes, but as far as I can tell this is not actually useful for livegrep since the index file data is only consumed exclusively by the codesearch process.

In livegrep, enabling eager memory load on the code search index tangibly reduces average search query latency due to minimizing major page fault overhead in the kernel. However, this comes at the cost of slightly higher reload latency, since the kernel needs to read ahead the entire index file to load it into the page table. For an index with ~450K files and ~100M lines of source, a hot reload without `MAP_POPULATE` takes about ~2 seconds while a hot reload with `MAP_POPULATE` takes about ~3 seconds.

Some real-world performance data from a livegrep cluster (which executed about ~50K queries over the time interval shown here):

Each spike in network ingress represents a download of a newly built index by automation (and subsequent hot reload). Without eager memory load:

before-network

before-page-faults

P50 search query execution latency reported by the server-side search timer:

before-latency

With eager memory load:

after-network

after-page-faults

after-latency

(This is another patch we've been using for several years that I forgot to upstream)

### DIFF
--- a/src/dump_load.cc
+++ b/src/dump_load.cc
@@ -25,6 +25,10 @@
 
 #include "google/protobuf/util/json_util.h"
 
+#include "gflags/gflags.h"
+
+DEFINE_bool(eager_memory_load, false, "Eagerly load memory-mapped index file pages into virtual memory (Linux only)");
+
 class codesearch_index {
 public:
     codesearch_index(code_searcher *cs, string path) :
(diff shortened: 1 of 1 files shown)


nelhage:
Oh, that's excellent. I'd been doing a manual preload on livegrep.com (literally `cat`ing the file to `/dev/null` 😂). This is much nicer.

PR #324, opened 2022-02-24
fix path for nginx conf in Docker image

xvandish:
Sorry I missed this! 

The extra `}` that snuck in was creating a seperate directory, like
```
livegrep/...
livegrep}/nginx.conf
```

So when nginx tried to start up with livegrep/nginx.conf it failed.


nelhage:
Whoops! Thanks for the fix :)

xvandish:
welcome! tks for taking a look :)

Issue #331, opened 2022-04-16
`index_time_` weirdness


xvandish:
Hello,

`index_time`, as returned by the `Info` gRPC call, is returning either `13` or `14` as the timestamp. The `Codesearch` backend that serves the index and the `indexer` the creates the Index runs are both GKE pods, running on Docker images with the exact same Dockerfiles as what's in the main repo.

Things I've checked:
1. I port forwarded the Codesearch pod to my local computer, and spun up the Go webserver, and confirmed that `/debug/stats` reflects the weird `13` or `14` number by reporting that the index is very old, even though it was actually just a few minutes old when I checked.  For example, `{"index_age":1650085072}` (time in seconds since index was created).
2. In Google's Cloud Shell, I installed the `grpc_cli`, then in another tab of the shell, port forwarded the Codesearch pod's port to local. Then I ran `./grpc_cli call localhost:8080 CodeSearch.Info ''` and got back
```
trees......
index_time: 14
```
3. I `kubectl exec`'d both into the Codesearch backend pod and into an instance of a Cron based indexer and ran `date +%s` and both times the output looked fine. For example `1650134782`. And using that same value in Go gets the following output.
```
num = 1650134782
t := time.Unix(num, 0)
fmt.Printf("%v\n", t)

Output:
2022-04-16 18:46:22 +0000 UTC
```

So It's seems like the clock on both machines is ok, not what's going wrong in [these](https://github.com/livegrep/livegrep/blob/main/src/codesearch.cc#L511-L513) lines.
```
timeval now;
gettimeofday(&now, NULL);
index_timestamp_ = now.tv_sec;
```

Any ideas what's going on? I'll keep looking into this, but figured someone might have an idea. 

Thanks!

xvandish:
Alright, more debugging -
TLDR - issue may be from using `stat.st_mtime` when loading the index

I added a few debug lines to [src/codesearch.cc](https://github.com/xvandish/livegrep/commit/09fe5dfc799d5731786c46df0fa41a9efecb5976)
```
index_timestamp_ = now.tv_sec;
+ fprintf(stdout, "now.tv_sec: %ld\n", now.tv_sec);
+ fprintf(stdout, "index_timestamp_: %lld\n", index_timestamp_);
```

and to [src/dump_load.cc](https://github.com/xvandish/livegrep/commit/008d91b7b9b15efaa0f65fc3ee574ad89be29a27)
```
[...]

PR #332, opened 2022-04-18
Fix stack overflow in inotify-based reload watcher routine

LINKIWI:
While working on the unrelated patch described https://github.com/livegrep/livegrep/issues/331#issuecomment-1101043856 I found a stack overflow bug in the inotify-based index reload watcher.

A `read` on the inotify fd returns a `struct inotify_event`, plus the name of the affected file, followed by a null termination character. The previous logic allowed a full read of all this data, but into only a container of size `struct inotify_event`. This change allocates a buffer with the maximum possible read length to avoid stack corruption.

To reproduce:

Build an index `/tmp/index`, and load it with:

```
$ ./bazel-bin/src/tools/codesearch --load_index /tmp/index --hot_index_reload
```

```
$ touch /tmp/index
```

Observe stack overflow:

```
Serving...
*** stack smashing detected ***: terminated
Aborted (core dumped)
```

With this patch:

```
Serving...
Detected change to index file; reloading...
Serving...
Detected change to index file; reloading...
Serving...
```

### DIFF
--- a/src/lib/fs_linux.cc
+++ b/src/lib/fs_linux.cc
@@ -11,6 +11,8 @@
 #include <sys/inotify.h>
 #include <unistd.h>
 
+const ssize_t max_event_len = sizeof(struct inotify_event) + NAME_MAX + 1;
+
 namespace {
     int fd = -1;
     int wd = -1;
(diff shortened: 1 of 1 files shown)


nelhage:
Oof, good find. Thanks for the fix.

PR #333, opened 2022-04-18
Encode index creation timestamp into dump file

### LINKED ISSUE #331: `index_time_` weirdness

LINKIWI:
This change proposes encoding the index timestamp (derived at index build time) directly into the index dump. This has a number of advantages:

* Removes the runtime dependency on the `stat(2)` syscall to derive the index timestamp, which on some environments/platforms/filesystems (?) can be unreliable or inaccurate. See #331.
* Allows for deterministic recovery of an index file, independent of filesystem attributes or state. In other words, as long as the byte contents of the index file are unchanged, the `Info()` RPC should always return the same response across invocations of `codesearch`.
* Is more semantically accurate--users are probably interested in the time at which the index was built, rather than the time it was deployed.

Unfortunately, since this is a backwards-incompatible change to the index file format, it requires bumping the index version.

Fixes #331.


src/dump_load.cc
```diff
@@ -309,6 +317,9 @@ void codesearch_index::dump_metadata() {
     hdr_.name_off = stream_.tellp();
     dump_string(cs_->name());
 
+    hdr_.timestamp_off = stream_.tellp();
+    dump_int64(cs_->index_timestamp());
```
nelhage, line 321: Seems to me it would make sense to just store the timestamp in the header, instead of using the offset pattern? For large blobs we store the offsets, but for scalars we can just put a `uint64_t timestamp` directly in the header and it should work fine.

nelhage (changes requested):
I'm +1 to this approach; Per the comment let's just store the timestamp inline in the header instead of as a random scalar hanging out later in the bytestream.

PR #336, opened 2022-05-31
Score and rank git files before indexing

xvandish:
Hello! 

This PR introduces a way to improve search results, by scoring and ranking files prior to indexing, so that more "annoying" files get indexed last and so generally show last in search results. 

The idea works really well! On both MB and GB scale corpora, the down-ranked files show at the bottom of search results, or don't show at all until `max_matches` is extended. Indexing performance is generally a bit faster than before, more on that below.

We take the following steps:
1. We walk all repos/files before indexing them. While walking, we give each file/blob a score.
2. We sort the list of all files based on score (and, in the multi-threaded case, by repo name as well)
3. We then walk the sorted list, and call `cs->index_file` to actually index the file.

## Performance 
We're now generally faster at indexing than before. The speedup is greater when there are more git repos to index, thanks to a multi-threaded filesystem walk described below. Most of the time the speedup is within a few seconds. Here's a comparison run:
```
$ (score-and-rank) ./bazel-bin/src/tools/codesearch -index-only -dump-index nelhage.idx ./repos/nelhage.json
...
repository indexed in 159.871761s
== begin metrics ==
index.bytes 1716122017
index.bytes.dedup 1421882081
index.content.chunks 113
index.content.ranges 38506740
index.data.chunks 11
index.files 124302
index.lines 44171401
index.lines.dedup 29077802
== end metrics ==

$ (main) ./bazel-bin/src/tools/codesearch -index-only -dump-index nelhage.idx ./repos/nelhage.json
...
repository indexed in 163.871761s
== begin metrics ==
index.bytes 1716122017
index.bytes.dedup 1421882081
index.content.chunks 113
index.content.ranges 38506740
index.data.chunks 11
index.files 124302
index.lines 44171401
index.lines.dedup 29077802
== end metrics ==
```

I've used `hyperfine` to test performance with multiple runs and with/without warmups and caches, and this assertion (we're now faster) generally holds.

## Multi-threading

We use multiple threads to walk the git trees within a repo. We don't have each thread tackle a single git repo, instead, while walking repos we push each root tree/directory to a `thread_queue` that some workers work off of. This turned out to be faster than each each thread taking a repo at a time. I'm fairly sure this is because workloads are now more balanced, instead of a single thread being stuck indexing a huge repo, multiple threads at a time can help out with said huge repos trees.

Users can opt out of multi-threading, and we ourselves opt out if there aren't enough repos to index, as generally the time it takes to open the threads dominates indexing time with small workloads. We also opt out if the `FLAGS_order_root` flag is on, since a specific ordering isn't really easy to do when walking with multiple threads.

## Memory
I've run this through valgrind and don't see any memory leaks.
```
==330891== LEAK SUMMARY:
==330891==    definitely lost: 0 bytes in 0 blocks
==330891==    indirectly lost: 0 bytes in 0 blocks
==330891==      possibly lost: 0 bytes in 0 blocks
==330891==    still reachable: 91,554 bytes in 1,287 blocks
==330891==         suppressed: 0 bytes in 0 blocks
```

The number of reachable blocks is consistent with valgrind output for previous indexer versions.

Also, as far as I can tell, memory usage is consistent with prior versions.

## Random thoughts
* I couldn't find a way I was happy with to positively boost files. There are comments in `src/score.cc` about that.
* I've done my best with logging, which got harder with having multiple threads. I'm pretty happy with where it ended up though, try an indexing run and let me know what you think.
* We're pretty IO bound now, since we're efficiently using a machines core's to walk repos, we're bound by the disk access speeds to read blobs and odb info.
* I've added functionality to bump the number of file descriptors the process can have open. This is pretty necessary on macOS, where the default soft limit is 256. Users can opt out. They can't opt in and specify a max, we auto bump to the systems hard limit. That might be a nice thing to add in.
* At the moment, users can opt out of threads completely, but can't choose thread count. Would be fairly easy to add in, might be nice
* Sorting search results like in #111 would be a great next thing, when/if this change lands.
* I didn't add functionality to opt out of scoring/ranking. Might be nice to add in.
* It was faster to use a mutex and push to the global list of files than for each thread to keep a local list and then push once it's done. That surprised me, I figured contention might be an issue, but I guess not.
* The most expensive thing we really do is `git_blob_lookup` and `git_tree_entry_to_object`. There's no way I can to keep the blobs in memory - so we don't have to do lookups twice - without ballooning our memory usage to a huge amount. When I tried keeping the blobs in memory, indexing was almost twice as fast, like 10mins vs 19mins on 5000 repos, so there may be a "high performance" version we can chase that keeps a bounded cache of git_blobs. Unbounded is probably a no go, on those 5000 repos codesearch was using like 50GB of ram during indexing.

This is my first time ever writing C++, so apologies for the rough spots!

Anything at all let me know, happy to change things around.

*note, build is failing atm (but shouldn't be), I think it's because I committed a few too many times within an hour and the connection to the remote cache has been hit with too many requests recently. Will re-trigger it tomorrow. On that note, on my fork I've got a system like your initial one using github caches, that works super well for a fast commit pace, I can upstream that if you'd like.

### DIFF
--- a/src/git_indexer.h
+++ b/src/git_indexer.h
@@ -37,7 +37,7 @@ public:
     void begin_indexing();
 protected:
     int get_next_repo_idx();
-    void process_repos(threadsafe_progress_indicator *tpi);
+    void process_repos(int estimatedReposToProcess, threadsafe_progress_indicator *tpi);
     void walk(git_repository *curr_repo,
             const std::string& ref,
             const std::string& repopath,
@@ -57,7 +57,6 @@ protected:
                    std::vector<std::unique_ptr<pre_indexed_file>>& results);
     void index_files();
     void print_last_git_err_and_exit(int err);
-    void walk_repositories_subset(int start, int end, threadsafe_progress_indicator *tpi);
 
     code_searcher *cs_;
     std::string submodule_prefix_;
--- a/src/git_indexer.h
+++ b/src/git_indexer.h
@@ -31,6 +31,8 @@ public:
     git_indexer(code_searcher *cs,
                 const google::protobuf::RepeatedPtrField<RepoSpec>& repositories);
     ~git_indexer();
+    void begin_indexing();
+protected:
     void walk(git_repository *curr_repo,
             const std::string& ref,
             const std::string repopath,
@@ -38,8 +40,6 @@ public:
             Metadata metadata,
             bool walk_submodules,
             const std::string& submodule_prefix);
-    void begin_indexing();
-protected:
     void walk_tree(const std::string& pfx,
                    const std::string& order,
                    const std::string repopath,
@@ -49,6 +49,7 @@ protected:
                    git_tree *tree,
                    git_repository *curr_repo);
     void index_files();
+    void print_last_git_err_and_exit(int err);
 
     code_searcher *cs_;
     std::string submodule_prefix_;
--- a/src/git_indexer.h
+++ b/src/git_indexer.h
@@ -25,6 +25,7 @@ struct pre_indexed_file {
     std::string  repopath;
     std::string  path;
     std::string id; // git oid
+    std::unique_ptr<git_oid> oid;
     int score;
     git_repository *repo;
 };
--- a/src/git_indexer.h
+++ b/src/git_indexer.h
@@ -11,8 +11,9 @@
 #include <string>
 #include "src/proto/config.pb.h"
 #include "src/smart_git.h"
-#include "src/lib/per_thread.h"
+/* #include "src/lib/per_thread.h" */
 #include "src/lib/threadsafe_progress_indicator.h"
+#include "src/lib/thread_queue.h"
 
 class code_searcher;
 class git_repository;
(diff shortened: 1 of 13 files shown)


src/git_indexer.h
```diff
@@ -10,34 +10,79 @@
...
+// used to walk a repos folders/trees in parallel using threads
+struct tree_to_walk {
+    int id;
+    std::string prefix;
+    std::string repopath;
+    bool walk_submodules;
+    string submodule_prefix;
+    const indexed_tree *idx_tree;
+    git_tree *tree;
+    git_repository *repo;
```
xvandish, line 45: We eat the space overhead of storing a pointer to a git repo with every tree and file, since even with millions of files and trees it's only 10s of MB, and it's slower (in experiments) than having git_repo in a map and then fetching them out when needed.

xvandish:
Converting to draft again, as I've noticed a few things in our deployment - 
1. Memory usage increased (prev peak of like 6GB to like 20GB) more than it had during local testing
2. The indexing progress bar that uses carriage returns works terrible in GKE

PR #340, opened 2022-06-20
Update mirrored git clones when remote HEAD changes

### LINKED ISSUE #293: Renaming default branch breaks livegrep-github-reindex / livegrep-fetch-reindex
From my observations, renaming the default branch on GitHub.com breaks repository indexing.

As an example, I created a new repository called `Glovo/test-repo-for-livegrep-indexing`. The default branch was originally called `main`.

Here is the clone created by livegrep:

```
➜  test-repo-for-livegrep-indexing git:(main) pwd
/var/livegrep/Glovo/test-repo-for-livegrep-indexing
➜  test-repo-for-livegrep-indexing git:(main) cat HEAD
ref: refs/heads/main
➜  test-repo-for-livegrep-indexing git:(main)
```

Everything gets indexed correctly at this stage.

But once I rename the default branch from `main` to `default`, the indexing doesn't work anymore.

Here's an excerpt from the logs of `livegrep-github-reindex`:

```
livegrep-indexer_1   | 2021-04-27T19:44:23.020534955Z   done
livegrep-indexer_1   | 2021-04-27T19:44:23.020653758Z Walking repo_spec name=Glovo/test-repo-for-livegrep-indexing, path=/data/Glovo/test-repo-for-livegrep-indexing (including  submodules: false)
livegrep-indexer_1   | 2021-04-27T19:44:23.021321667Z   walking HEAD
livegrep-indexer_1   | 2021-04-27T19:44:23.022362821Z ref HEAD not found, skipping (empty repo?)
livegrep-indexer_1   | 2021-04-27T19:44:23.022378007Z   done
```

And here is the local clone of the repo:

```
➜  test-repo-for-livegrep-indexing git:(main) pwd
/var/livegrep/Glovo/test-repo-for-livegrep-indexing
➜  test-repo-for-livegrep-indexing git:(main) cat HEAD
ref: refs/heads/main
```

Note that the `HEAD` file stays at `main`.

Once I delete that directory and inspect it again, after another run of the indexer, it's correct now:

```
➜  test-repo-for-livegrep-indexing git:(default) pwd
/var/livegrep/Glovo/test-repo-for-livegrep-indexing
➜  test-repo-for-livegrep-indexing git:(default) cat HEAD
ref: refs/heads/default
```

It also gets correctly included in the index. I'll try to see if that's something we can fix ourselves, but I want to report the issue already.

xvandish:
Hello - this patch attempts to solve sync issues between the mirrored clones that `livegrep-github-reindex` creates and their upstream counterparts - described in #293. TLDR of that issue is that when changes are made to the HEAD refs of the remotes of locally cloned mirror repos,  `git fetch -p` will invalidate the current local HEAD, but won't update it to match the new remote HEAD, which  eventually leads livegrep to think that repo is empty when it's indexing, which causes it to be skipped.

To fix this, we make an extra call to git - `git ls-remote --symref origin HEAD`. We then compare the remote `HEAD` information to the local `HEAD`, and if different, then we manually update with `git symbolic-ref HEAD {newHead}`.

Even with the extra call to git, this patch doesn't slow down indexing runs. Updating even 5000 repos takes exactly the same amount of time it used to, thanks to the goroutines used for the `git fetch` and `git ls-remote` calls.

To test this, I ran `fetch-reindex`, then modified the default branch name for a few of the repos, then re-ran `fetch-reindex` and made sure that the patch picked up the modifications to the upstream. Example run:

```
$ bazel-bin/cmd/livegrep-fetch-reindex/livegrep-fetch-reindex_/livegrep-fetch-reindex -out=xvandish.idx repos/xvandish.json
$ // then I went to github and renamed the default branch of livegrep-fragment from `good_main_5` to `good_main_6`.
$ bazel-bin/cmd/livegrep-fetch-reindex/livegrep-fetch-reindex_/livegrep-fetch-reindex -out=xvandish.idx repos/xvandish.json
  ...
  From github.com:xvandish/livegrep-fragment
  - [deleted]         (none)     -> good_main_5
  * [new branch]      good_main_6 -> good_main_6
  xvandish/livegrep-fragment: remote HEAD: refs/heads/good_main_6 does not match local HEAD: refs/heads/good_main_5. Attempting to fix...
  xvandish/livegrep-fragment: HEAD update done.
  ...
```

I've also been running this on our company instance, and its successfully found and fixed the instances of this problem. 

Some random thoughts:
* eventually, `git remote set-head origin -a` may be able to do this (`ls-remote..` + `symbolic-ref..`) by itself. A patched version of git enables this, but who knows if/when it would make it upstream. See [this thread on the git mailing list I started](https://public-inbox.org/git/CANWRddMXrhs2grTfq9NtHM7KqxaFRejvaXxvDGV90_m=Kmm-uA@mail.gmail.com/) for more info. 
* I gated this functionality behind a (by default enabled) flag in the spirit of not making unexpected changes - but since this theoretically leads to more correct deployments than before, I think it also makes sense to not gate it at all.
* I modified `callGit` so that it could output to either `stdout/err` or a buffer, which adds some complexity. I'm not a huge fan of it, but I went with that over a copy paste `callGit2` version that does the exact same thing as `callGit` but prints to a buffer. Happy to change that if you prefer.

This closes #293 .

### DIFF
--- a/cmd/livegrep-fetch-reindex/main.go
+++ b/cmd/livegrep-fetch-reindex/main.go
@@ -32,7 +32,7 @@ var (
 	flagStatsdOn      = flag.Bool("send-metrics-to-statsd", false, "Send indexing metrics to StatsD")
 	flagStatsdAddr    = flag.String("statsd-address", "", "address URI of statsd listener for metrics export")
 	flagStatsdPrefix  = flag.String("statsd-prefix", "", "optional prefix to apply to all metrics")
-	flagUpdateHead    = flag.Bool("update-head", true, "update the HEAD reference if it has changed on remote")
+	flagUpdateHead    = flag.Bool("update-head", true, "update the local HEAD ref if it is different than remote. Done when updating already cloned repos.")
 )
 
 // Used to extract the refname from a line like the following:
--- a/cmd/livegrep-fetch-reindex/main.go
+++ b/cmd/livegrep-fetch-reindex/main.go
@@ -179,28 +179,77 @@ if test "$1" = "get"; then
 fi
 `)
 
+// runs exec.Output() and returns the command output
+func callGit2(program string, args []string, username string, password string) ([]byte, error) {
+	var err error
+	var out []byte
+
+	if username != "" || password != "" {
+		f, err := passwordHelper(username, password, &args)
+		if err != nil {
+			return nil, err
+		}
+		defer os.Remove(f.Name())
+	}
+
+	for i := 0; i < 3; i++ {
+		cmd := exec.Command("git", args...)
+		if password != "" {
+			r, w, err := os.Pipe()
+			if err != nil {
+				return nil, err
+			}
+
+			cmd.ExtraFiles = []*os.File{r}
+
+			go func() {
+				defer w.Close()
+				w.WriteString(password)
+			}()
+			defer r.Close()
+		}
+		if username != "" {
+			cmd.Env = append(os.Environ(), fmt.Sprintf("LIVEGREP_GITHUB_USERNAME=%s", username))
+		}
+		out, err = cmd.Output()
+		if err == nil {
+			return out, nil
+		}
+	}
+	return nil, fmt.Errorf("%s %v: %s", program, args, err.Error())
+}
+
+func passwordHelper(username string, password string, args *[]string) (*os.File, error) {
+	// If we're given credentials, pass them to git via a
+	// credential helper
+	//
+	// In order to avoid the key hitting the
+	// filesystem, we pass the actual key via a
+	// pipe on fd `3`, and we set the askpass
+	// script to a tiny sh script that just reads
+	// from that pipe.
+	f, err := ioutil.TempFile("", "livegrep-credential-helper")
+	if err != nil {
+		return nil, err
+	}
+	f.WriteString(credentialHelperScript)
+	f.Close()
+
+	os.Chmod(f.Name(), 0700)
+	*args = append([]string{"-c", fmt.Sprintf("credential.helper=%s", f.Name())}, *args...)
+
+	return f, nil
+}
+
 func callGit(program string, args []string, username string, password string) error {
 	var err error
 
 	if username != "" || password != "" {
-		// If we're given credentials, pass them to git via a
-		// credential helper
-		//
-		// In order to avoid the key hitting the
-		// filesystem, we pass the actual key via a
-		// pipe on fd `3`, and we set the askpass
-		// script to a tiny sh script that just reads
-		// from that pipe.
-		f, err := ioutil.TempFile("", "livegrep-credential-helper")
+		f, err := passwordHelper(username, password, &args)
 		if err != nil {
 			return err
 		}
-		f.WriteString(credentialHelperScript)
-		f.Close()
 		defer os.Remove(f.Name())
-
-		os.Chmod(f.Name(), 0700)
-		args = append([]string{"-c", fmt.Sprintf("credential.helper=%s", f.Name())}, args...)
 	}
 
 	for i := 0; i < 3; i++ {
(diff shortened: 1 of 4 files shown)


cmd/livegrep-fetch-reindex/main.go
```diff
@@ -26,8 +28,17 @@ var (
 	flagReloadBackend = flag.String("reload-backend", "", "Backend to send a Reload RPC to")
 	flagNumWorkers    = flag.Int("num-workers", 8, "Number of workers used to update repositories")
 	flagNoIndex       = flag.Bool("no-index", false, "Skip indexing after fetching")
+	flagMetricsPath   = flag.String("metrics-out", "", "File that indexing metrics should be sent to")
```
nelhage, line 31: Looks like these were left and probably not intended to be part of this PR?

cmd/livegrep-fetch-reindex/main.go
```diff
@@ -26,8 +28,17 @@ var (
...
+	flagStatsdOn      = flag.Bool("send-metrics-to-statsd", false, "Send indexing metrics to StatsD")
+	flagStatsdAddr    = flag.String("statsd-address", "", "address URI of statsd listener for metrics export")
+	flagStatsdPrefix  = flag.String("statsd-prefix", "", "optional prefix to apply to all metrics")
+	flagUpdateHead    = flag.Bool("update-head", true, "update the local HEAD ref if it is different than remote. Done when updating already cloned repos.")
```
nelhage, line 35: Let's just do this unconditionally, I think the new behavior is just correct and there's no need to add the flag unless there are known sharp edges.

cmd/livegrep-fetch-reindex/main.go
```diff
@@ -166,8 +177,12 @@ if test "$1" = "get"; then
 fi
 `)
 
-func callGit(program string, args []string, username string, password string) error {
+// calls cmd.Run() if returnOutput is false
+// and cmd.Output() otherwise
+// always returns an out []byte, but it will always be nill if returnOutput is false
+func callGit(program string, args []string, username string, password string, returnOutput bool) ([]byte, error) {
```
nelhage, line 183: Let's provide two wrappers around this that give a nicer interface than the bool and maybe-`nil` output. I'm thinking about Python's `check_call`/`check_output` distinction here for some inspiration.

We might call them

```
func callGit(program string, args []string, username, password string) error
func callGitGetOutput(program string, args []string, username, password string) ([]byte, error)
```

This will make call sites much clearer.

We can keep one implementation -- rename the existing function to `callGitInternal` or something, and then pass appropriate values for `output` from the wrappers.

cmd/livegrep-fetch-reindex/main.go
```diff
@@ -260,7 +283,77 @@ func checkoutOne(r *config.RepoSpec) error {
...
+	// See https://public-inbox.org/git/CANWRddPDhM1g6rtu-a2a=EogXD_hOFwSDsgMCbVvB7dibMaEqw@mail.gmail.com/T/#t
+	// for confirmation on this approach from the Git folks.
+	//
+	// To update the HEAD ref we do the following:
+	// 1. Get the remote head ref			- (git ls-remote --symref origin HEAD)
+	// 2. Get the current local head ref	- (git symbolic-ref HEAD)
+	// 3. Compare them. If outdated, update the local to match remote - (git symbolic-ref HEAD new_ref)
+	// We use goroutines to call `git fetch -p` and `git ls-remote --symref origin HEAD` in "parallel"
+	// becase they each take ~1.5s.
+	wg := sync.WaitGroup{}
```
nelhage, line 305: Let's use an [`errgroup`](https://pkg.go.dev/golang.org/x/sync/errgroup) here, which will help with the "merging the errors" for us and avoid the error-prone `wg.Add(2)`.

nelhage (changes requested):
Thanks for the fix! I think this approach makes sense and I'd suggest we just do it unconditionally.

I left a number of comments on the implementation -- overall it looks fine but I had a few suggestions that I think will clean it up a bit.

cmd/livegrep-fetch-reindex/main.go
xvandish: ah, good catch, removed.

cmd/livegrep-fetch-reindex/main.go
xvandish: Sounds good. I haven't run into any edges so far, will send a patch if I do.

cmd/livegrep-fetch-reindex/main.go
```diff
@@ -260,7 +290,66 @@ func checkoutOne(r *config.RepoSpec) error {
...
+	// See https://public-inbox.org/git/CANWRddPDhM1g6rtu-a2a=EogXD_hOFwSDsgMCbVvB7dibMaEqw@mail.gmail.com/T/#t
+	// for confirmation on this approach from the Git folks.
+	//
+	// To update the HEAD ref we do the following:
+	// 1. Get the remote head ref			- (git ls-remote --symref origin HEAD)
+	// 2. Get the current local head ref	- (git symbolic-ref HEAD)
+	// 3. Compare them. If outdated, update the local to match remote - (git symbolic-ref HEAD new_ref)
+	// We use goroutines to call `git fetch -p` and `git ls-remote --symref origin HEAD` in "parallel"
+	// becase they each take ~1.5s.
+	g := new(errgroup.Group)
```
nelhage, line 306:
```suggestion
	var g errgroup.Group
```

cmd/livegrep-fetch-reindex/main.go
xvandish: Nice, I like that. I've gone ahead and implemented this pretty much exactly as you described, works great.

cmd/livegrep-fetch-reindex/main.go
```diff
@@ -166,8 +173,24 @@ if test "$1" = "get"; then
...
+
+// calls a `git ...` command. Output is added to a buffer and returned
+func callGetGetOutput(program string, args []string, username string, password string) ([]byte, error) {
+	buff, err := callGitInternal(program, args, username, password, true)
+	return buff, err
+}
+
+// calls cmd.Run() if returnOutput is false
+// and cmd.Output() otherwise
+// always returns an out []byte, but it will always be nill if returnOutput is false
```
nelhage, line 190:
```suggestion
// always returns an out []byte, but it will always be nil if returnOutput is false
```

cmd/livegrep-fetch-reindex/main.go
xvandish: Sounds good, this is done now.

nelhage (approved):
Thanks for the fixes!

xvandish:
> Thanks for the fixes!

Welcome, thanks for the review!

PR #343, opened 2022-07-30
Make /search API requests with POST

### LINKED ISSUE #259: Select All results in HTTP 414
We have ~400 repos, and clicking Select All leaves the UI in a weird state that looks unresponsive. Worse yet, there seems to be a memory of the last repos you selected so users basically permanently break themselves.

Is there some mitigation we can put in for it somewhere?

image

isker:
With GET, the `repos` parameter may grow the URL beyond the maximum
length permitted by the server.

Use POST instead.

Fixes #259.

Also, improve the bazel installation instructions in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,10 @@ can see a running instance at
 Building
 --------
 
-livegrep builds using [bazel][bazel]. You will need to
-[install][bazel-install] a fairly recent version: as of this writing
-we test on bazel 4.0.0.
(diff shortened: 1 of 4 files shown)


isker:
I tested this commit locally, both with and without the JS change, to verify backward compatibility.

nelhage (approved):
Looks great! Thanks for the contribution.

PR #346, opened 2022-08-03
Support darkmode with CSS media query

clint-stripe:
## Summary
Use CSS media queries to define custom properties ('variables') for the colors used throughout the livegrep UI. There were _quite a few_ colors used previously, so even in the light mode variant this is a little bit different from the current CSS. I simplified some colors to be the same, rather than the slight differences they had before.

I threw a little deployment of this PR on https://lg.clint64.net/search/livegrep, feel free to poke around. (Your browser can simulate light/dark mode preferences with the dev tools -- in Chrome, open the "Rendering" tool (probably hit Escape to open that extra tool drawer??), in Firefox these are icons in a sidebar of the "Inspector" pane.)

I'm not at all tied to any particular colors here, and I may have missed some places that need additional styling! 

## Future work

I think we might _eventually_ want to add an explicit toggle for this in the UI, but that requires a little bit more work: there's no way to take the media query preference _or_ override it without duplicating the entire block of colors (:css-intensifies:) from what I can tell.

A rough sketch of that future state, I think:

```css
:root, :root[data-color-theme=light] {
  --color-background: #fff;
}

:root[data-color-theme=dark] {
  --color-background: #000;
}

@media (prefers-color-scheme: dark) {
  :root {
    --color-background: #000;
  }
}
```

Light mode, either explicitly or by fallback, is defined once. Dark mode is defined both in the media query block _and_ outside it, to handle the explicit and 'auto' case. If you know of a better solution here, I would love to hear about it!

### DIFF
--- a/web/htdocs/assets/css/codesearch.css
+++ b/web/htdocs/assets/css/codesearch.css
@@ -1,14 +1,160 @@
+:root {
+    --color-background: #ffffff;
+    --color-background-subtle: #f8f8f8;
+    --color-background-hover: #d4d4d4;
+    --color-background-modal-overlay: rgba(0, 0, 0, 0.1);
+    --color-foreground: #000000;
+    --color-foreground-muted: #404040;
+    --color-foreground-subtle: #7d7d7d;
+    --color-foreground-emphasis: #000000;
+    --color-shadow: rgba(0, 0, 0, 0.25);
+
+    --color-foreground-matchstr: inherit;
+    --color-background-matchstr: rgba(255, 234, 170, 0.75);
+    --color-outline-highlight-focus: rgba(255, 234, 170, 0.5);
+    --color-background-highlight-focus: rgba(255, 234, 170, 0.33);
+
+    --color-foreground-symlink: rgba(60, 60, 60);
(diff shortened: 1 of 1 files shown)


clint-stripe:
Oops, I forgot the screenshots 🙂 


lg clint64 net_search_livegrep_q=fast fold_case=auto regex=false context=true (5)

lg clint64 net_search_livegrep_q=fast fold_case=auto regex=false context=true (4)

nelhage:
This looks great, thanks!

Also, gosh, CSS variables. Pretty sure we didn't even have those yet when we first wrote this code :)

isker:
> there's no way to take the media query preference or override it without duplicating the entire block of colors (:css-intensifies:) from what I can tell.

Having seen people that actually understand CSS (unlike myself 🌞) implement a dark mode at work, I think this is correct.  We have taken the toil out of it at build time, however, with this:

https://github.com/csstools/postcss-extend-rule

```css
.darkMode {
  /* vars go here */
}

@media (prefers-color-scheme: dark) {
  @extend .darkMode
}
```

PR #347, opened 2022-08-04
Fix example url_pattern

isker:
When not using the file viewer, metadata.github takes precedence over
metadata.url_pattern.  When using the file viewer, only url_pattern is
consulted.  url_pattern is incorrect for this repository.  Fix it.


nelhage:
Thanks!

PR #354, opened 2022-12-03
Allow to click on query hint

afourneau:
Hello, this PR allow to click on the query hint to add it to the search and then focus on the search input to complete the query hint.


nelhage:
Neat, this makes sense to me!

I wonder if it should append to the query string instead of replacing? I can imagine being surprised if I was typing some query and tried to add `path:` by clicking, and it ate my entire query. What do you think?

nelhage:
Oh, whoops, I was misreading the code -- it looks like that's exactly what it already does!

PR #356, opened 2023-01-05
Use rules_js instead of org_dropbox_rules_node

clintharrison:
I was a bit bored over the holidays and thought this would be a low-risk way to slightly modernize the JS build for livegrep.

https://github.com/aspect-build/rules_js is the current community-supported way to build JS with Bazel; the `bazel_build_rules_nodejs` successor, essentially.

Technically I didn't convert the lockfiles _exactly_ -- there doesn't appear to be a way to create a single lockfile by merging the three `npm-shrinkwrap.json` files, and so I just let the resolution happen again with the same major/minor versions of the top-level packages we care about: webpack, style-loader, css-loader.

These rules also don't have the same concept of an "internal npm package", so I opted for the simple but workable approach: a webpack resolver alias for `html` -> `3d/html`.

No real strong opinion on actually landing this, but since I got it working, I figured I'd share :)


nelhage:
Wow, this looks great! I've been wanting to get us off that ancient rules_nodejs. I'll take a look but I expect that I'll happily merge this.

nelhage:
Thank you again!

PR #358, opened 2023-02-06
add tls and no verify

yogeshg:


server/backend.go
```diff
@@ -32,7 +36,13 @@ type Backend struct {
 }
 
 func NewBackend(id string, addr string) (*Backend, error) {
-	client, err := grpc.Dial(addr, grpc.WithInsecure())
+	creds := insecure.NewCredentials()
+	if val := os.Getenv("TLS_SKIP_VERIFY"); val != "" {
+		creds = credentials.NewTLS(&tls.Config{
```
yogeshg, line 41: tls is required when the index-reading-backend is setup as an independent (cloud run) service on gcs.

nelhage:
@yogeshg Can you take a look at the failing tests?

yogeshg:
> @yogeshg Can you take a look at the failing tests?

Of course!
I checked and it looks like I need to include credentials/ and credentials/insecure in bazel BUILD file. I am trying to run .circleci/build.sh in the docker container built from .circleci/Dockerfile with bazel_version changed to 5.3.2. I am stuck at a step where it's trying to download boost. I have included the best guess I have for these files, so it would be nice to have that tested through CI, but looks like that awaits your permission. Please lmk what's the best way to proceed here.

yogeshg:
@nelhage bump in case this got burried :)

> Of course! I checked and it looks like I need to include credentials/ and credentials/insecure in bazel BUILD file. I am trying to run .circleci/build.sh in the docker container built from .circleci/Dockerfile with bazel_version changed to 5.3.2. I am stuck at a step where it's trying to download boost. I have included the best guess I have for these files, so it would be nice to have that tested through CI, but looks like that awaits your permission. Please lmk what's the best way to proceed here.

yogeshg:
Closing since github search is very powerful now.

PR #361, opened 2023-02-28
Allow Go server config to specify grpc dial opts

### LINKED ISSUE #360: livegrep frontend doesn't support non-default MaxCallRecvMsgSize
#321 added flags to change the recv/send message size on the `codesearch` backend, but larger-than-default sizes will still be rejected by the `livegrep` side.

We probably want an option in config.json for this, and update `server/backend.go` accordingly to pass along `grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize())`

clint-stripe:
Fixes #360.

I deployed this change and with the corresponding flag to `codesearch` was actually able to see larger responses in the frontend. (Something like `max_matches:3000` is enough to hit the limit for the query I ran.)

The `WORKSPACE` change is just to silence a warning.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -139,6 +139,7 @@ load("@aspect_rules_js//npm:npm_import.bzl", "npm_translate_lock")
 
 npm_translate_lock(
     name = "npm",
+    npmrc = "//web:.npmrc",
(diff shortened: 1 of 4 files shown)


nelhage:
Nice, thanks for the PR.

Issue #326, opened 2022-03-03
SPA SSR considerations


xvandish:
Hi @nelhage ! 

I was running into some performance issues with large `max_matches` values (like in #94) so I figured I'd try writing the UI in a more modern framework, to see if it could solve the perf pain points. 

I've started work on the UI as a Svelte rendered app ([SvelteKit](https://kit.svelte.dev/)), and good news - it solves the perf issues and is a super nice experience! Unfortunately, moving to any sort of SPA framework means we no longer get SSR as we do now, which seems a huge shame to lose. Maybe I'm getting too tied up in losing SSR, but if feels really nice to use websites when pages load pretty much instantly, without huge shifts in content, and it's so nice compared to the state of a lot of the web.

What do you think, is transitioning to the SPA without SSR worth the tradeoff? 
Pros:
1. Performance wins in render times pretty much everywhere
2. Much smaller bundle sizes - no need for backbone, jquery, bootstrap, etc
4. Dev experience for the UI itself is much nicer (hot-module reloading with Vite). 
5. Easy to serve, same as now the `/` handler is served with the FileHandler (or Nginx does it)

Cons:
1. No SSR!! Users will see blank pages until the JS loads then some API call is made made (this is mostly on the fileviewer / directory explorer). Issue can be mitigated by pre-loading pages, and it works pretty well, but this only helps after the initial load.

I thought a bit about what the architecture would look like if we did try to get SSR, and from what I can see, any sort of non-brittle SSR JS experience involves a Node server somewhere, and I can't think of a way to do it without introducing complexity or completely changing things. Here's a diagram of what I'm envisioning
```
[SPA] --- /api or /someAsset -> [Node server] ----- proxy /api, api calls for SSR --> [Go server]
```
Random thoughts: 
* `cmd/livegrep.go` should probably spin up both the node and go servers. 
* The Node server should be able to function without a reverse proxy in front of it, so it should itself proxy `/api` requests to the Go backend. 
* Any functionality within the Go api server that's proxy or header related needs to be replicated/shifted to the Node server...
* We now have to add Node to the docker images 
* It's just one more process/thing to have to consider -_-
* On the completely changing things, rewriting the entire server to be in Node is possible, but just need to make sure the Node grpc functionality is seamless. Can Bazel even compile proto to Nodejs atm?

[...]

PR #363, opened 2023-03-30
Expose search page configuration in API endpoint

clint-stripe:
The list of repo URLs is useful to have exposed outside the templated HTML, along with any custom link configs: I'm poking at some experimental frontend/CLI changes and it would be helpful to have access to this configuration.

We only have one backend -- I'm not sure how widely used multiple backends are? Would that be potentially useful too?

(I _think_ something similar was done in the changes discussed in #326 but that wasn't upstreamed and I'm not seeing any [similar routes in their current fork](https://github.com/xvandish/livegrep/blob/my_main/server/server.go#L607).)

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -80,10 +80,17 @@ func (s *server) ServeRoot(ctx context.Context, w http.ResponseWriter, r *http.R
 	http.Redirect(w, r, "/search", 303)
 }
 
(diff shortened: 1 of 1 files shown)


nelhage (approved):
This makes sense to me!

I'm not sure that anyone is using multiple backends. I wrote it for livegrep.com at some point but haven't used it in that configuration for a long time. Seems fine to omit for now.

PR #366, opened 2023-04-16
git_indexer: Gracefully handle git tree entries that cannot be converted to objects

LINKIWI:
This PR should fix a `codesearch` crash in the rare (?) event that a git tree entry fails to convert to a git object for further analysis in the `walk_tree` logic path, by ensuring that `git_tree_entry_to_object` returns `0` before extracting the object contents.

This fixes the following crash I encountered while indexing a large repository:

```
codesearch: external/com_github_libgit2/src/object.c:219: git_object_type: Assertion `obj' failed.
```

I didn't investigate the underlying root cause of the error. It could be due to a corrupted git tree, or a tree entry unrecognized by the current version of libgit2, or something else.

Manual verification, doing offline indexing of several repositories:

```
...
Walking repo_spec name=private/repo, path=/home/ubuntu/.cache/livegrep/repository/TOA7hwyWPTPulYLHyoAdPGw4CRy2AT22oPbI2whDXac= (including  submodules: false)
  walking origin/master
  done
Walking repo_spec name=private/repo, path=/home/ubuntu/.cache/livegrep/repository/lpr-RoHEQyBUdPifRIdtD0k2WMB2FuI0uL_aBHFnCPY= (including  submodules: false)
  walking origin/master
Unable to convert git tree entry 46225f9851b671b55a91e3e4a61d3651593f95ba to object, skipping
  done
Walking repo_spec name=private/repo, path=/home/ubuntu/.cache/livegrep/repository/REF-YHsenVA4mcPpggdLDGMFnax1Qmuv_RIFgcMuPng= (including  submodules: false)
  walking origin/master
  done
...
```

### DIFF
--- a/src/git_indexer.cc
+++ b/src/git_indexer.cc
@@ -83,7 +83,16 @@ void git_indexer::walk_tree(const string& pfx,
     for (vector<const git_tree_entry *>::iterator it = ordered.begin();
          it != ordered.end(); ++it) {
         smart_object<git_object> obj;
-        git_tree_entry_to_object(obj, repo_, *it);
+        char oid[GIT_OID_HEXSZ + 1];
+
+        const git_oid* entry_oid = git_tree_entry_id(*it);
+        git_oid_tostr(oid, GIT_OID_HEXSZ + 1, entry_oid);
+
+        if (git_tree_entry_to_object(obj, repo_, *it) != 0) {
+            fprintf(stderr, "Unable to convert git tree entry %s to object, skipping\n", oid);
+            continue;
+        }
+
         string path = pfx + git_tree_entry_name(*it);
 
         if (git_tree_entry_type(*it) == GIT_OBJ_TREE) {
@@ -108,11 +117,7 @@ void git_indexer::walk_tree(const string& pfx,
             git_indexer sub_indexer(cs_, sub_repopath, string(sub_name), meta, walk_submodules_);
             sub_indexer.submodule_prefix_ = submodule_prefix_ + path + "/";
 
-            const git_oid* rev = git_tree_entry_id(*it);
-            char revstr[GIT_OID_HEXSZ + 1];
-            git_oid_tostr(revstr, GIT_OID_HEXSZ + 1, rev);
-
-            sub_indexer.walk(string(revstr));
+            sub_indexer.walk(string(oid));
         }
     }
 }


nelhage:
Seems reasonable enough to me. Doesn't seem like something that should happen normally but I can imagine various weird reasons it might.

asutherland:
It seems like this probably broke git submodule indexing.  The libgit2 docs don't really provide any clarity about this, but it seems like [git_tree_entry_to_object calls git_object_lookup](https://github.com/libgit2/libgit2/blob/a9aa21476e34679fb7e381d04356c7a94bc698fe/src/libgit2/tree.c#L296) which [just calls git_object_lookup_prefix](https://github.com/libgit2/libgit2/blob/a9aa21476e34679fb7e381d04356c7a94bc698fe/src/libgit2/object.c#L284) which [ends up calling git_odb_read if not cached](https://github.com/libgit2/libgit2/blob/a9aa21476e34679fb7e381d04356c7a94bc698fe/src/libgit2/object.c#L251).  But it makes sense that this would fail because the commit isn't in the current repo/tree, it's in the submodule repo/tree.

The manual verification in the root comment shows `including  submodules: false` which I think means submodules weren't tested [based on the code that emits that printf](https://github.com/livegrep/livegrep/blob/52fa4423cc879d01f5f8e098af2c1395c5388572/src/tools/codesearch.cc#L112).

asutherland:
Reverting the commit made our git submodules be indexed again, and the log output seemed to line up with that (logs in [our searchfox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1924797)).

It shouldn't be too hard to change things so that the early return to avoid the assert can still happen and submodules can still work and I can try and provide a patch for that.  Would https://github.com/livegrep/livegrep/blob/main/client/test/integration_test.go or something derived from it be the right way to try and add test coverage for submodules?

nelhage:
> Would https://github.com/livegrep/livegrep/blob/main/client/test/integration_test.go or something derived from it […]
Something derived from that is probably workable. That specific file is a bit heavyweight and does fairly end-to-end integration on the kernel (which is of course a pretty large repo), but it should be possible to do something similar for a smaller test case, and get end-to-end verification that we're working right.

Potentially, I would check in a script that **generates** a test-case repository in a tempdir using the CLI, and then indexes that, in order to avoid downloading an external target? That's potentially a sweet spot that keeps the source of truth relatively readable/editable. But I wouldn't be too unhappy if we did something like check in a zipfile or a tarball containing a small test repo making use of submodules, or similar. I think doing that feels preferable to fetching a remote during CI, but I'm not certain.

PR #369, opened 2023-06-01
Add config option for file extension, language mapping

clint-stripe:
I've _long_ wanted a way to add new file extensions without a PR, merge, build dance -- finally got around to adding it.

The specific motivation was terraform `.tf` files, which a newer version of PrismJS includes (as generic HCL, Hashicorp configuration language). I updated the example, though maybe we do want that as a default inclusion too.

Edit: I threw another small commit here to use a regex on the first line too -- we often have small shell (and Ruby of course 🙂) scripts that have no extension, but nevertheless would be much nicer to have highlighted:

`bin/groups-graph`:
image

### DIFF
--- a/doc/examples/livegrep/server.json
+++ b/doc/examples/livegrep/server.json
@@ -9,5 +9,8 @@
     "feedback": {
         "mailto": "nelhage@nelhage.com"
     },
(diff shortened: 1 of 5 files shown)


nelhage (approved):
Very nice, I like it.

PR #370, opened 2023-06-22
Add repos grouping by their path.

ivanchgeek:
Added repos grouping by their name path. Example:
image

Also added an ability to select/unselect all repos in group by click on it's header.

### DIFF
--- a/web/src/codesearch/repo_selector.js
+++ b/web/src/codesearch/repo_selector.js
@@ -10,6 +10,17 @@ function init() {
         liveSearch: true,
         width: '20em'
     });
+    $('#repos').on('refreshed.bs.select', function () {
+        var headers = $(this).parent().find('.dropdown-header');
+        headers.css('cursor', 'pointer');
+        headers.on('click', function (event) {
+            event.stopPropagation();
+            var optgroup = $('#repos optgroup[label="' + $(this).text() + '"]')
+            var allSelected = !optgroup.children('option:not(:selected)').length;
+            optgroup.children().prop('selected', !allSelected);
+            $("#repos").selectpicker('refresh');
+        });
+    });
     $(window).on('keyup', '.bootstrap-select .bs-searchbox input', function(event) {
         var keycode = (event.keyCode ? event.keyCode : event.which);
         if(keycode == '13'){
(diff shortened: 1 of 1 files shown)


ObjatieGroba:
Cool feature

amdei:
This is exactly what the doctor ordered!
Most suitable quality of life improvement ever!

In our setup we have 30+ projects in different groups/subgroups, and clicking all of them in specific group is a nightmare...
With this feature personally my life become happier.

web/src/codesearch/repo_selector.js
```diff
@@ -35,11 +46,17 @@ function updateOptions(newOptions) {
     if (_.isEqual(currentOptions, newOptions)) {
         return;
     }
-
+
```
amdei, line 49: Seems extra spaces should be removed.

amdei:
BTW: it would be nice to have repo names sorted in lexicographic order, no in order they was added to admin interface.

web/src/codesearch/repo_selector.js
```diff
@@ -37,9 +48,16 @@ function updateOptions(newOptions) {
...
 
     $('#repos').empty();
+    newOptions.sort()
+
     for (var i = 0; i < newOptions.length; i++) {
-        var option = newOptions[i];
-        $('#repos').append($('<option>').attr('value', option).text(option));
+        var path = newOptions[i].split('/');
+        var group = path.slice(0, path.length - 1).join('/') + '/';
+        var groupQuery = '#repos' + (path.length == 1 ? '' : ' optgroup[data-path="' + group + '"]');
```
nelhage, line 56: I have some FUD about using a DOM query here; It's probably fine but relying on the DOM and jQuery selectors to do our grouping is going to be way less efficient (and, IMO, less clear) than just grouping *first* and directly appending the elements we want.

Can we instead use [`_.groupBy`](https://underscorejs.org/#groupBy) to first group by the name, and then just loop over those groups and directly append the `<optgroup>` and `<option>` tags we want? I think I'd expect that to be clearer to read and also a bit faster.

ivanchgeek: I think I found a good compromise. I'd tried to use _.groupBy, but in this approach we need to 
* separate group name and repo name
* group then
* convert result object in array using Object.entries to easily iterate it (also we need sort it again by group name, because this method can crush groups order)
Therefore, much more code is generated and the clearness stay under big question. The result of my try is here:
`newOptions.sort()
    var separatedGroupOptions = newOptions
    .map(function (option) {
        var path = option.split('/');
        return {
            group: path.slice(0, path.length - 1).join('/') + '/', 
            repo: path[path.length - 1]
        }
    });
    var groupedOptions = Object.entries(_.groupBy(separatedGroupOptions, 'group'))
    .map(function (group) {
        var groupName = group[0]
        var clearGroupOptions = group[1].map(function (option) { return option.repo })
        return [groupName, clearGroupOptions]
    });

    for (var i = 0; i < groupedOptions.length; i++) {
        var group = groupedOptions[i][0];
        var options = groupedOptions[i][1];
        var groupDOM = group == '/' ? $('#repos') : $('<optgroup>').attr('label', group)
        for (var j = 0; j < options.length; j++) {
            var option = options[j]
            groupDOM.append($('<option>').attr('value', group + option).text(option))
        }
        if (group != '/') $('#repos').append(groupDOM)
    }
`

Commited way to do it avoids excess dom quering, but save initial locality and readability.

nelhage:
Thanks for the PR!

PR #371, opened 2023-07-04
Include git in base dockerfile

SoMuchForSubtlety:
The base image currently only spits out "Error reading file". Adding git to the image fixes the issue.


nelhage:
Hm. This was missing because it's not necessary if we're only using the index, and not the file-view. But `git` is fairly small and it seems fine to me to add it to all the images.

PR #374, opened 2023-07-19
Fix search update loss, when selecting repos using group header.

ivanchgeek:
In recently added repos grouping feature bug was detected. After selecting a number of repos using group header, search results weren't updating. Small fix for this problem here.


nelhage:
Thanks!

PR #376, opened 2023-09-14
Add an "I'm feeling lucky" query param

clint-stripe:
Someone mentioned to me they'd like to be able to jump directly the file viewer (or on GitHub etc.) when there's only a single result -- useful for linking from Slack or documentation for a unique constant, for example.

(That said this _is_ a little bit silly. It's also not discoverable because I didn't add it to the UI. (For use in a go link, it's fine.) It also ended up being a bit more code than I hoped, but :go-intensifies:. Feel free to merge, or not!)

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


nelhage:
Hm. Did you consider implementing this in the frontend? We could load the full page and do the normal query, and *then* trigger the redirect in Javascript if `lucky=1`. Probably be harder to avoid a flash of livegrep UI, but I'd guess it'd be much less code and also require less keeping-in-sync between the Go and the Javascript.

clintharrison:
Yeah, I went back and forth a bit before deciding the backend approach was a little easier to understand overall -- you can take a look at the other version in https://github.com/livegrep/livegrep/commit/bc367ca66a8de5c441e6212c24fceec32e503ad4. This might also be my complete unfamiliarity with Backbone speaking though :)
That commit alone isn't ideal given the lack of "search in progress" indicator (or any warning that you're about to be redirected away) 😕 

(I _think_ we could also see inconsistent results depending on users' preferences, if something like `fold_case` or `regex` isn't in the URL, and they fall back to the user's preferences. But this feature is only ever a 'nice-to-have' anyway.)

PR #380, opened 2023-10-05
Support queries with multiple path: terms

### LINKED ISSUE #378: Support for multiple file: or path: terms
I often write queries that already make use of `path:foo` -- but the suggested "Narrow to filetype" buttons adds a `file:\.rb$` which is then an error.

Is there any fud with allowing multiple? I put together [a WIP attempt](https://github.com/clint-stripe/livegrep/commit/b70ed14326b4096a6fac69a655b6204d65e7694e) which...is pretty invasive :) It still needs a bit of cleanup and some additional tests if we think this seems reasonable at all 😆

clintharrison:
This PR fixes #378.

(This is probably easiest to review commit-by-commit. The first one was necessary to get anything building on the Macs I have access to, and the second isn't necessary but makes editing Bazel-y C++ and Go much more pleasant.)

Changing the `Query` proto to accept a repeated `file` (or `not_file`) is straightforward, with one slightly odd caveat. `FilenameOnly` queries did not make use of the `file` field but, for some reason, repurposed `line`. This is still true, but the `line` pattern is used as the "most important" one, since we don't support multiple bounds in a match (see https://github.com/livegrep/livegrep/pull/349).

The changes to _use_ the additional values are mostly boring loops where previously one check sufficed. The tagsearch changes might just be incorrect, though 😆

This doesn't include any frontend changes, but in the future we may want to "fix" the conditional rendering of the extension buttons to allow `path:` too (the [current heuristic](https://github.com/livegrep/livegrep/blob/7e86d05914b84199d25ebab964df2935d436db1e/web/src/codesearch/codesearch_ui.js#L614C8-L616) hides it more often than necessary).

<details>
<summary><h4>Example searches</h4></summary>

`path:c build`

image

<hr>

Clicking the `.cc` button: `file:.cc path:c build`

image
</details>

### DIFF
--- a/server/query.go
+++ b/server/query.go
@@ -38,10 +38,20 @@ func onlyOneSynonym(ops map[string]string, op1 string, op2 string) (string, erro
 	return ops[op2], nil
 }
 
+func ensureSingleValue(ops map[string][]string, key string) (string, error) {
+	if len(ops[key]) > 1 {
+		return "", fmt.Errorf("multiple values for %s:", key)
+	}
+	if len(ops[key]) == 1 {
+		return ops[key][0], nil
+	}
+	return "", nil
+}
+
 func ParseQuery(query string, globalRegex bool) (pb.Query, error) {
 	var out pb.Query
 
-	ops := make(map[string]string)
+	ops := make(map[string][]string)
 	key := ""
 	term := ""
 	q := strings.TrimSpace(query)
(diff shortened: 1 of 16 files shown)


server/query.go
```diff
@@ -146,20 +147,39 @@ func ParseQuery(query string, globalRegex bool) (pb.Query, error) {
...
 
+	// This is a special case to provide a better error message,
+	// since the main search term is represented by the "" op.
+	if len(ops[""]) > 1 {
+		return out, fmt.Errorf("main search term must be contiguous")
+	}
+
+	// Handle synonyms
+	out.File = append(ops["file"], ops["path"]...)
+	out.NotFile = append(ops["-file"], ops["-path"]...)
```
clintharrison, line 158: This _does_ mean we might reorder terms, and the one that ends up in `Line` for a `FilenameOnly` search might not be the first one. I don't know if this matters that much, though :P

nelhage (approved):
Nice! Looks fairly clean -- thanks for the easily-reviewed PR.

PR #382, opened 2023-11-16
ignore_symlinks option

wfrisch:
Add an option to ignore symbolic links, off by default.

Some projects contain infinite symlink loops ([Example](https://github.com/jbruchon/jdupes/tree/master/testdir/recursed_a)) or symlinks that point outside the directory tree. The former just slows down indexing because the recursion stops after a certain maximum depth, while the latter can be a security issue in certain setups.

There are workarounds that don't require changing livegrep, such as a [filtering read-only FUSE](https://github.com/wfrisch/rofs-filtered), but IMHO it would be more comfortable to just have this option in livegrep itself.


nelhage:
Yeah, this seems reasonable.

PR #386, opened 2024-03-08
Adding new GA4 tagging code snippets to common layout

yasrivera:
## Context

Livegrep is still with the Google Analytics 3 mechanism of collecting data from the website according to the customer's ID. Since Google Analytics 4 was launched and the previous was deprecated, the data is not being collected properly. We must add the new method provided by GA platform in order to have analytics data being collected properly and sent to the customer's Google Analytics dashboard.

## Solution

I noticed that Livegrep uses Google Tag Manager instead of the basic method provided by Google Analytics 4. After some research, I found this [blog post](https://www.analyticsmania.com/post/how-to-install-google-tag-manager/) where we can identify that method provided by Google Tag Manager.

First, we have to change the previous implemented function inside the `<head>` tag to the new one. Then, we need to add a `<noscript>` tag right after the `<body>` tag opening.

### DIFF
--- a/web/templates/common/layout.html
+++ b/web/templates/common/layout.html
@@ -35,19 +35,16 @@
     {{if .Config.GoogleAnalyticsId}}
     <script{{.Nonce}} type="text/javascript">
 
-      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-      })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
-
-      ga('create', '{{.Config.GoogleAnalyticsId}}', 'auto');
-      ga('send', 'pageview');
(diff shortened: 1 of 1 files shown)


nelhage:
Looks good to me -- I see the PR is marked as Draft, should I hold off on merging for any reason?

yasrivera:
@nelhage I haven't tested yet (build `livegrep` and check if it renders properly), that's why I marked as Draft

nelhage:
Sounds good! Drop a note when you have a chance to test and I'll merge. I've switched off GA personally so I don't have a quick testbed set up.

yasrivera:
Hey @nelhage! I wasn't able to test it properly. I'm following what's said in readme till [this part](https://www.google.com/search?q=Excessive%20heat&hl=en-GB&stick=H4sIAAAAAAAAAONQUeLVT9c3NCwrskipMDcxf8Tozi3w8sc9YSm7XWcbqy25NLhQFSiJc4nqJ-fn5KQml2Tm5-mnZBYnFpekFhULMEpxc3Hq5-obJBeYJqfxLGLlc61ITi0uzixLVchITSwBAAnRfPNsAAAA) and even adding an example of google analytics tag in `server.json` file.

I noticed that everything which is placed inside a block like `{{if .Config.<variable-from-config>}}` doesn't actually appear when building locally. I'm not sure if `livegrep` is using a specific framework that dynamically insert these values or if I should do something else when building/running it.

Would you have any guess about this issue?

nelhage:
It's been a while since I had this in my head, unfortunately. I'd expect putting something into the `server.json` would suffice but I'm not certain if there's something subtle happening.

yasrivera:
I see... Then, do you think it's safe to merge this change related only to updating the Google Analytics mechanism even though it wasn't tested? E.g, the `feedback` option in the frontend isn't broken. Probably it's something related to the local building, in the production it seems fine.

nelhage:
Thanks for the fix!

yasrivera:
Nice! Thank you for the support @nelhage! :D

PR #387, opened 2024-03-14
Remove Stripe brand from opensearch.xml

sirodoht:
This PR removes the Stripe keyword from the `<link rel=search>` tag for `opensearch.xml`.


nelhage:
Heh, good catch. Thanks!

PR #389, opened 2024-04-25
Make window.scriptData initialization hashable

stefanc-stripe:
This PR makes it possible to serve Livegrep content using a CSP policy that relies on hashes without using Nonce.

Previously, it was not possible to create a hash of the inline script that consumes `{{.ScriptData}}` because the data is dynamically generated and the hash is not stable. This inline script was the only `<script>` tag for which a hash cannot be computed.

With this PR, it's possible to compute the hash for the inline script because the dynamic data is stored in a harmless `application/json` block while the JS script tag has stable content.


nelhage:
Makes sense to me. Thanks for the fix!

Issue #390, opened 2024-06-12
Build fails - ModuleNotFoundError: No module named 'six.moves'


MarPur:
I'm trying to build the repository on a fresh Ubuntu 24.04 installation and the build fails

```
$ bazelisk build //... --sandbox_debug
DEBUG: Rule 'com_github_nelhage_rules_boost' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1683368611 +0000"
DEBUG: Repository com_github_nelhage_rules_boost instantiated at:
  /home/martynas/Downloads/livegrep/WORKSPACE:28:15: in <toplevel>
Repository rule git_repository defined at:
  /home/martynas/.cache/bazel/_bazel_martynas/b845a7b6a5adbf465acf11e26bc6e767/external/bazel_tools/tools/build_defs/repo/git.bzl:199:33: in <toplevel>
DEBUG: /home/martynas/.cache/bazel/_bazel_martynas/b845a7b6a5adbf465acf11e26bc6e767/external/bazel_gazelle/internal/go_repository.bzl:209:18: org_golang_x_net: gazelle: finding module path for import golang.org/x/sys/windows: finding module path for import golang.org/x/sys/windows: package golang.org/x/sys/windows: build constraints exclude all Go files in /home/martynas/.cache/bazel/_bazel_martynas/b845a7b6a5adbf465acf11e26bc6e767/external/bazel_gazelle_go_repository_cache/pkg/mod/golang.org/x/sys@v0.21.0/windows
gazelle: finding module path for import golang.org/x/sys/windows: finding module path for import golang.org/x/sys/windows: package golang.org/x/sys/windows: build constraints exclude all Go files in /home/martynas/.cache/bazel/_bazel_martynas/b845a7b6a5adbf465acf11e26bc6e767/external/bazel_gazelle_go_repository_cache/pkg/mod/golang.org/x/sys@v0.21.0/windows
gazelle: finding module path for import golang.org/x/sys/windows: finding module path for import golang.org/x/sys/windows: package golang.org/x/sys/windows: build constraints exclude all Go files in /home/martynas/.cache/bazel/_bazel_martynas/b845a7b6a5adbf465acf11e26bc6e767/external/bazel_gazelle_go_repository_cache/pkg/mod/golang.org/x/sys@v0.21.0/windows
[...]

PR #391, opened 2024-06-13
Use pkg_tar from rules_pkg

MarPur:
Add `rules_pkg` to use `pkg_tar` instead of using `pkg_tar` from `bazel_tools`, as this seems to fix [issue](https://github.com/livegrep/livegrep/issues/390) I reported about build not working on Ubuntu 24.04.

Including `rules_pkg` is also required to upgrade to later versions of Bazel (Bazel 6 and higher), since it was moved into its own repository.


nelhage:
Thanks!

Issue #400, opened 2025-03-27
Build issues in macOS


martynasp-wix:
This is the error I get when I try to build macOS Sequoia `15.3.2`

```
ERROR: /private/var/tmp/_bazel_martynasp/82954a5b1a1dda88f5a959810d54d1f4/external/upb/BUILD:80:11: Compiling upb/upb.c failed: (Exit 1): cc_wrapper.sh failed: error executing command external/local_config_cc/cc_wrapper.sh -U_FORTIFY_SOURCE -fstack-protector -Wall -Wthread-safety -Wself-assign -Wunused-but-set-parameter -Wno-free-nonheap-object -fcolor-diagnostics ... (remaining 25 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
external/upb/upb/upb.c:223:25: error: defining a type within '__builtin_offsetof' is a Clang extension [-Werror,-Wgnu-offsetof-extensions]
  223 |   n = UPB_ALIGN_DOWN(n, UPB_ALIGN_OF(upb_arena));
      |                         ^~~~~~~~~~~~~~~~~~~~~~~
external/upb/upb/port_def.inc:94:38: note: expanded from macro 'UPB_ALIGN_OF'
   94 | #define UPB_ALIGN_OF(type) offsetof (struct { char c; type member; }, member)
      |                                      ^~~~~~
/Library/Developer/CommandLineTools/usr/lib/clang/16/include/__stddef_offsetof.h:16:43: note: expanded from macro 'offsetof'
   16 | #define offsetof(t, d) __builtin_offsetof(t, d)
      |                                           ^
external/upb/upb/port_def.inc:92:48: note: expanded from macro 'UPB_ALIGN_DOWN'
   92 | #define UPB_ALIGN_DOWN(size, align) ((size) / (align) * (align))
      |                                                ^~~~~
external/upb/upb/upb.c:223:25: error: defining a type within '__builtin_offsetof' is a Clang extension [-Werror,-Wgnu-offsetof-extensions]
  223 |   n = UPB_ALIGN_DOWN(n, UPB_ALIGN_OF(upb_arena));
      |                         ^~~~~~~~~~~~~~~~~~~~~~~
external/upb/upb/port_def.inc:94:38: note: expanded from macro 'UPB_ALIGN_OF'
   94 | #define UPB_ALIGN_OF(type) offsetof (struct { char c; type member; }, member)
      |                                      ^~~~~~
[...]

PR #401, opened 2025-06-05
Upgrade Bazel version to 6.5.0

MarPur:
Bazel version 5 is deprecated and no longer supported, while Bazel 6 is still in maintenance mode until end of 2025 [^1].

I'm making this for two reasons:
* _Maybe_ this will be enough to address https://github.com/livegrep/livegrep/issues/400 - `upb` was merged into `grpc` [^2] (I have not checked this on macos yet).
* If not, then I guess the technically correct solution would be to configure a toolchain to use a specific compiler version instead of relying on host's compiler version, but I suspect that would require a newer version of Bazel (e.g. [toolchains_llvm](https://github.com/bazel-contrib/toolchains_llvm) requires at least Bazel 6).

If I'm wrong, then bumping Bazel's version can't hurt. 

[^1]: https://bazel.build/release
[^2]: https://github.com/protocolbuffers/upb


MarPur:
Looks like this PR needs more attention - it doesn't build on macos arm

PR #402, opened 2025-06-19
fix listen property format in docker run command

dave-cz:


nelhage:
The change seems fine for consistency, but just for clarity, is the current version actually broken? I believe that should work with go's `flag` package…

dave-cz:
With `=` in command I get error

`open .0.0.0:8910: no such file or directory`

without it works properly

